PR #5, opened 2016-05-23
bump scalajs to 0.6.9

grzesiul:


ghik:
ScalaJS is already 0.6.9

grzesiul:
Indirectly, monifu 1.2 requires 0.6.8, so I didn't checked the latest version

ghik:
Are these versions incompatible? I think you could just force version in SBT and be fine.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.15.0"
\ No newline at end of file
+version in ThisBuild := "1.16.0"
```
ghik: If ScalaJS versions are compatible, this should be just `1.15.1`

PR #8, opened 2016-06-03
ScalaInterop

mjjaniec:
We have more and more Scala/Java mixed code. Sometimes it will be useful to have ability to convert collections to/from scala code in Java Code


ghik:
From Java you can use `scala.collection.JavaConversions`. Is it lacking something?

mjjaniec:
Your right. Shame on me...

PR #11, opened 2016-07-13
Try equivalent for Future's traverse and sequence

MilusW:


commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -77,13 +78,13 @@ object SharedExtensions extends SharedExtensions {
       * Maps a `Future` using [[concurrent.RunNowEC RunNowEC]].
       */
     def mapNow[B](f: A => B): Future[B] =
-      fut.map(f)(RunNowEC)
+    fut.map(f)(RunNowEC)
```
mkej: What? Why? This should be indented, as well as the line below.

commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -128,4 +129,43 @@ object SharedExtensions extends SharedExtensions {
     def toNOpt: NOpt[A] =
       if (tr.isFailure) NOpt.Empty else NOpt.some(tr.get)
   }
+
+  object TryOps {
```
ghik: I don't think `SharedExtensions` is the right place for it. `SharedExtensions` is intended to keep extension methods that can be inherited by package object (e.g. `package object ump`) so that these extension methods are visible "globally".

ghik:
Oh, maybe this is the time to use [`cats`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala) or [`scalaz`](https://github.com/scalaz/scalaz/blob/scalaz-seven/core/src/main/scala/scalaz/Traverse.scala)? 

Nah, just kidding.

ghik:
How about doing something akin to https://github.com/scala/scala/pull/5233/commits/ea1d370506cd3f12f1a0bda49f1316f98123ef4d ?

PR #12, opened 2016-08-15
NamedEnum::name should be a val.

Crossy147:


commons-core/src/test/scala/com/avsystem/commons/misc/NamedEnumTest.scala
```diff
@@ -0,0 +1,36 @@
...
+class NamedEnumTest extends FunSuite {
+  sealed trait SomeNamedEnum extends NamedEnum
+  object SomeNamedEnum extends NamedEnumCompanion[SomeNamedEnum] {
+
+    case object First extends SomeNamedEnum {
+      override def name: String = "First"
+    }
+
+    case object Second extends SomeNamedEnum {
+      override def name: String = "Second"
```
ddworak: You might want to check if this works with all possible ways to override `name`, e.g.: `val`, `lazy val`, inheriting name from a trait or abstract class. I'm also not sure, neither after reading the docs nor after reading this test, what happens if I call a method in `name` impl.

PR #16, opened 2016-11-09
Generated GenCodec should not skip unknown fields

Starzu:


ghik:
Skipping is necessary for retaining backwards compatibility - it allows for safe removal of case class fields.

PR #19, opened 2017-02-16
CollectionOps fixes

ddworak:
* rename unusable `toMap` 
* use `breakOut`

### DIFF
--- a/commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
+++ b/commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
@@ -206,12 +206,8 @@ object SharedExtensions extends SharedExtensions {
   }
 
   class CollectionOps[C[X] <: BTraversable[X], A](private val coll: C[A]) extends AnyVal {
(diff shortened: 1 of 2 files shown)


commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -206,27 +206,23 @@ object SharedExtensions extends SharedExtensions {
   }
 
   class CollectionOps[C[X] <: BTraversable[X], A](private val coll: C[A]) extends AnyVal {
-    def toMap[K, V](keyFun: A => K, valueFun: A => V): Map[K, V] = {
-      val res = Map.newBuilder[K, V]
-      coll.foreach { a =>
-        res += ((keyFun(a), valueFun(a)))
-      }
-      res.result()
+    def asMap[K, V](keyFun: A => K, valueFun: A => V): Map[K, V] = {
```
ghik: I understand why this must be done, but `asMap` kind of suggest that this returns a view. Maybe `collectToMap` or `mkMap` would be better?

ddworak:
`mkMap` it is then, changes pushed

PR #20, opened 2017-03-14
forEmpty Option extension method

mjjaniec:


commons-shared/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -118,12 +118,26 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with Seria
...
     if (isEmpty) Right(right) else Left(value)
 
-  override def toString =
+  /**
+    * Apply side effect only if Opt is empty. It's a bit like foreach for Opt.Empty
+    * @param sideEffect - code to be executed if opt is empty
+    * @return the same opt
+    * @example {{{captionOpt.forEmpty(logger.warn("caption is empty")).foreach(setCaption)}}}
+    */
+  @inline def forEmpty(sideEffect: => Unit): Opt[A] = {
```
ghik: For consistency, we should have the same thing in `NOpt`, `OptRef` and `OptArg`

mjjaniec: right

PR #21, opened 2017-03-31
Relax most traversable ops constraints to TraversableOnce.

ddworak:


commons-shared/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -345,11 +340,26 @@ object SharedExtensions extends SharedExtensions {
     def containsAll(other: BTraversable[A]): Boolean = other.forall(set.contains)
   }
 
+  class TraversableOps[C[X] <: BTraversable[X], A](private val coll: C[A]) extends AnyVal {
+
+    def headOpt: Opt[A] = if (coll.isEmpty) Opt.Empty else Opt(coll.head)
+
+    def lastOpt: Opt[A] = if (coll.isEmpty) Opt.Empty else Opt(coll.last)
+
+    def mkMap[K, V](keyFun: A => K, valueFun: A => V): Map[K, V] = {
```
ghik: This could also be in `TraversableOnceOps` if you use `foreach` instead of `map` (like it was done before).

ddworak: Right, fixed now.

ghik:
This will be released as 1.19.15. Be careful when updating - this is a binary incompatible change, so make sure your dependencies are also updated.

PR #24, opened 2017-07-19
Added result type metadata.

sikor:
Now RPCMetadata contains also type class for RPC method result type.


ghik:
Cool, thanks for doing what I was too lazy to do.

PR #25, opened 2017-08-22
Mongo match elem

Passarinho4:
Could you check in? I think we should add elemMatch here too.


commons-mongo/src/main/scala/com/avsystem/commons/mongo/Filter.scala
```diff
@@ -42,6 +42,8 @@ object Filter {
 
   def notExists(key: DocKey[_, _]): Bson = F.exists(key.key, false)
 
+  def elemMatch(key: DocKey[_, _], filter: Bson): Bson = F.elemMatch(key.key, filter)
```
mkej: I think you could limit this to accept `key: DocKey[_, BsonArray]`. It might be tricky for users to create a proper `filter`, but it'll work.

PR #26, opened 2017-08-28
added "contains" method to mongo filtering utils

nmarcin92:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/Filter.scala
```diff
@@ -42,7 +44,11 @@ object Filter {
 
   def notExists(key: DocKey[_, _]): Bson = F.exists(key.key, false)
 
-  def elemMatch(key: DocKey[_, _], filter: Bson): Bson = F.elemMatch(key.key, filter)
+  def elemMatch(key: DocKey[_, _ <: BsonArray], filter: Bson): Bson = F.elemMatch(key.key, filter)
+
+  def contains[A, COL <: Iterable[A]](key: DocKey[COL, _ <: BsonArray], value: A)
+    (implicit cbf: CanBuildFrom[Nothing, A, COL]): Bson =
+    F.eq(key.key, key.codec.toBson((cbf() += value).result()).asScala.head)
```
ghik: I guess `asScala.head` could be replaced with `get(0)` to avoid wrapping.

PR #27, opened 2017-09-05
Added putOpt to Doc

Ugon:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/Doc.scala
```diff
@@ -22,6 +22,11 @@ class Doc(private val doc: BsonDocument) extends AnyVal {
     case _ => this
   }
 
+  def putOpt[A](key: DocKey[A, _ <: BsonValue], optValue: Opt[A]): Doc = optValue match {
```
ghik: It's a matter of taste, but this can be shorter: `optValue.fold(this)(put(key, _))`

PR #29, opened 2017-09-05
Codec Input/Output implementations for Bson Reader/Writer + BsonRef

mkej:
- added codec Input/Output implementations for Bson Reader/Writer
- added BsonRef for referencing Bson document attributes

### DIFF
--- a/commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
+++ b/commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
@@ -1,0 +1,83 @@
+package com.avsystem.commons
+package mongo
+
+import com.avsystem.commons.serialization.{ListOutput, ObjectOutput, Output}
+import org.bson.{BsonArray, BsonBinary, BsonBoolean, BsonDocument, BsonDouble, BsonInt32, BsonInt64, BsonNull, BsonString, BsonValue}
+
+class BsonValueOutput extends Output {
+  private var _value: Opt[BsonValue] = Opt.empty
+
+  private def setValue(bsonValue: BsonValue): Unit = {
+    _value match {
+      case Opt.Empty =>
+        _value = Opt.some(bsonValue)
+      case Opt(oldValue) =>
+        throw new IllegalStateException(s"Cannot set value to $bsonValue, value is already present: $oldValue")
+    }
+  }
+
+  def value: BsonValue = {
+    _value match {
+      case Opt.Empty =>
+        throw new IllegalStateException("Value is not set")
+      case Opt(bsonValue) =>
+        bsonValue
+    }
+  }
+
+  override def writeNull(): Unit = setValue(BsonNull.VALUE)
+  override def writeString(str: String): Unit = setValue(new BsonString(str))
+  override def writeBoolean(boolean: Boolean): Unit = setValue(BsonBoolean.valueOf(boolean))
+  override def writeInt(int: Int): Unit = setValue(new BsonInt32(int))
+  override def writeLong(long: Long): Unit = setValue(new BsonInt64(long))
+  override def writeDouble(double: Double): Unit = setValue(new BsonDouble(double))
+  override def writeBinary(binary: Array[Byte]): Unit = setValue(new BsonBinary(binary))
+  override def writeList(): ListOutput = new BsonValueListOutput(setValue)
+  override def writeObject(): ObjectOutput = new BsonValueObjectOutput(setValue)
+}
+
+class BsonValueListOutput(receiver: BsonArray => Unit) extends ListOutput {
+  private var lastOutput: Opt[BsonValueOutput] = Opt.empty
+  private val array = new BsonArray()
+
+  private def flushLast(): Unit = {
+    lastOutput.foreach(out => array.add(out.value))
+  }
+
+  override def writeElement(): Output = {
+    flushLast()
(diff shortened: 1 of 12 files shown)


commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
```diff
@@ -0,0 +1,83 @@
...
+  private def setValue(bsonValue: BsonValue): Unit = {
+    _value match {
+      case Opt.Empty =>
+        _value = Opt.some(bsonValue)
+      case Opt(oldValue) =>
+        throw new IllegalStateException(s"Cannot set value to $bsonValue, value is already present: $oldValue")
+    }
+  }
+
+  def value: BsonValue = {
```
ghik: That's just `_value.getOrElse(throw ...)`

commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
```diff
@@ -0,0 +1,83 @@
...
+  override def writeBinary(binary: Array[Byte]): Unit = setValue(new BsonBinary(binary))
+  override def writeList(): ListOutput = new BsonValueListOutput(setValue)
+  override def writeObject(): ObjectOutput = new BsonValueObjectOutput(setValue)
+}
+
+class BsonValueListOutput(receiver: BsonArray => Unit) extends ListOutput {
+  private var lastOutput: Opt[BsonValueOutput] = Opt.empty
+  private val array = new BsonArray()
+
+  private def flushLast(): Unit = {
```
ghik: Another, possibly cleaner approach is to make `BsonValueOutput` accept a consumer for its value. This way it can either write it to a variable or put directly into `BsonArray` or `BsonDocument`

commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
mkej: Yep, will change.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
```diff
@@ -0,0 +1,83 @@
+package com.avsystem.commons
+package mongo
+
+import com.avsystem.commons.serialization.{ListOutput, ObjectOutput, Output}
+import org.bson.{BsonArray, BsonBinary, BsonBoolean, BsonDocument, BsonDouble, BsonInt32, BsonInt64, BsonNull, BsonString, BsonValue}
+
+class BsonValueOutput extends Output {
+  private var _value: Opt[BsonValue] = Opt.empty
+
+  private def setValue(bsonValue: BsonValue): Unit = {
```
ghik: Superfluousss bracesss

commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
mkej: Will do.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala
mkej: I like to use braces whenever I have to use line breaks, even for a single expression.

commons-mongo/src/test/scala/com/avsystem/commons/mongo/BsonInputOutputTest.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import java.nio.ByteBuffer
+
+import org.bson._
+import org.bson.io.BasicOutputBuffer
+import org.scalatest.FunSuite
+
+class BsonInputOutputTest extends FunSuite {
+  test("Roundtrip binary encoding of random objects") {
+    100 times {
```
ghik: 100 times random with simple assert - this just _asks_ to be a property based test (like [this one](https://github.com/AVSystem/scala-commons/blob/master/commons-redis/src/test/scala/com/avsystem/commons/redis/protocol/RedisMsgTest.scala#L22) for instance)

commons-mongo/src/test/scala/com/avsystem/commons/mongo/Something.scala
```diff
@@ -0,0 +1,95 @@
+package com.avsystem.commons
+package mongo
+
+import com.avsystem.commons.serialization.GenCodec
+
+import scala.util.Random
+import scala.util.hashing.MurmurHash3
+
+object Randoms {
```
ghik: I guess `scalacheck` is not invented here :)

commons-mongo/src/test/scala/com/avsystem/commons/mongo/BsonInputOutputTest.scala
mkej: Of course, my dude - this test wast just moved verbatim from another codebase, where I didn't have such goodies.

commons-mongo/src/test/scala/com/avsystem/commons/mongo/Something.scala
mkej: See above, 10/10 would use `scalacheck`.

PR #30, opened 2017-09-13
Added codec for ObjectId

mkej:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonGenCodecs.scala
```diff
@@ -0,0 +1,17 @@
...
+import com.avsystem.commons.serialization.GenCodec
+import org.bson.types.ObjectId
+
+trait BsonGenCodecs {
+  implicit def objectIdCodec: GenCodec[ObjectId] = BsonGenCodecs.objectIdCodec
+}
+
+object BsonGenCodecs {
+  implicit val objectIdCodec: GenCodec[ObjectId] = GenCodec.createNullSafe(
+    readFun = _.asInstanceOf[BsonInput].readObjectId(),
```
ghik: Hmm, what I had in mind was not unconditionally assuming that `Input` is `BsonInput` but rather check if it is and fallback to some default behavior if it isn't.

mkej: Well... it would be really sad to silently receive a `String` where `ObjectId` was expected, but I guess it makes sense. Serializing to JSON will be possible then.

ghik: I don't see how this would be possible assuming that input used to deserialize is consistent with output used to serialize.

PR #33, opened 2017-09-18
Mongo 3.5.0 update + minor cleanup

mkej:
- updated driver to 3.5.0
- made specific driver depenencies optional
- moved driver-related utilities to separate packages


mkej:
Ignore the failed test, please - will be changed on #31.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/scala/GenCodecCollection.scala
```diff
@@ -0,0 +1,13 @@
+package com.avsystem.commons
+package mongo.scala
```
ghik: Please do not create packages named `scala`. This way we won't need all this `_root_` bullshit.

PR #37, opened 2017-09-28
Commons Kafka

Passarinho4:
Hi, I've added generic tools for serialization and deserialization events. Could you take a look?

### DIFF
--- a/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
+++ b/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
@@ -1,0 +1,26 @@
+package com.avsystem.commons
+package kafka
+
+import org.apache.kafka.common.serialization.{Deserializer, Serializer}
+
+import java.util
+
+import org.apache.kafka.common.serialization.Serde
+
+abstract class AbstractSerde[T] extends Serde[T] {
+  override def configure(configs: util.Map[String, _], isKey: Boolean): Unit = {}
+
+  override def close(): Unit = {}
+}
+
+abstract class AbstractSerializer[T] extends Serializer[T] {
+  override def configure(configs: util.Map[String, _], isKey: Boolean): Unit = {}
+
+  override def close(): Unit = {}
+}
+
+abstract class AbstractDeserializer[T] extends Deserializer[T] {
+  override def configure(configs: util.Map[String, _], isKey: Boolean): Unit = {}
+
+  override def close(): Unit = {}
+}
--- a/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
+++ b/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
@@ -3,8 +3,6 @@ package kafka
 
 import org.apache.kafka.common.serialization.{Deserializer, Serializer}
 
-import com.avsystem.commons.jiop.JavaInterop._
-
 import org.apache.kafka.common.serialization.Serde
 
 abstract class AbstractSerde[T] extends Serde[T] {
--- a/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
+++ b/commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
@@ -3,24 +3,24 @@ package kafka
 
 import org.apache.kafka.common.serialization.{Deserializer, Serializer}
 
-import java.util
+import com.avsystem.commons.jiop.JavaInterop._
 
 import org.apache.kafka.common.serialization.Serde
 
 abstract class AbstractSerde[T] extends Serde[T] {
-  override def configure(configs: util.Map[String, _], isKey: Boolean): Unit = {}
+  override def configure(configs: JMap[String, _], isKey: Boolean): Unit = {}
 
   override def close(): Unit = {}
 }
 
 abstract class AbstractSerializer[T] extends Serializer[T] {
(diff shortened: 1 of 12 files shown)


commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
```diff
@@ -0,0 +1,26 @@
+package com.avsystem.commons
+package kafka
+
+import org.apache.kafka.common.serialization.{Deserializer, Serializer}
+
+import java.util
```
ghik: Please don't. Import `JavaInterop._` and use `JMap`, `JList`, etc.

commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import org.apache.kafka.common.serialization.Serde
```
ghik: Why the hell didn't they call it `Deser` or `Sedes`?

commons-kafka/src/main/scala/com/avsystem/commons/kafka/SerdeRegistry.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import com.avsystem.commons.kafka.exceptions.UnsupportedVersionEvent
+
+import scala.collection.mutable
+
+
+class SerdeRegistry[T] extends AbstractDeserializer[T] {
+  private val registry = mutable.Map[Byte, KafkaSerde[_ <: T]]()
+
+  def add[X <: T](version: Byte, serde: KafkaSerde[X])(implicit classTag: ClassTag[X]): SerdeRegistry[T] = {
```
ghik: `classTag` looks unused.

BTW1 general tip: when a method is "fluent" and returns `this`, it's good to give it a return type `this.type` - this way it will work even if it's moved to a superclass or trait.

BTW2 why not add a convenience method that directly takes a `GenCodec`, e.g.

```scala
def add[X <: T](version: Byte)(implicit codec: GenCodec[T]): this.type = {
  registry += (version -> new KafkaSerde(codec))
  this
}
```

commons-kafka/src/main/scala/com/avsystem/commons/kafka/AbstractSerde.scala
Passarinho4: import org.apache.kafka.common.serialization.{Serde => Sedes} ? :P

commons-kafka/src/main/scala/com/avsystem/commons/kafka/VersionedSerializer.scala
```diff
@@ -0,0 +1,14 @@
+package com.avsystem.commons
+package kafka
+
+import com.avsystem.commons.kafka.exceptions.UnsupportedVersionEvent
+
+class VersionedSerializer[T](version: Byte)(implicit registry: SerdeRegistry[T]) extends AbstractSerializer[T] {
+  require(registry.isSupported(version), s"Version: $version isn't supported by registry. Please add it first.")
+  override def serialize(topic: String, data: T): Array[Byte] = {
+    registry(version) match {
+      case Some(serde: KafkaSerde[T]) => version +: serde.serializer().serialize(topic, data)
```
ghik: Ugh, that's going to copy the entire array... can we avoid it?

commons-kafka/src/main/scala/com/avsystem/commons/kafka/SerdeRegistry.scala
```diff
@@ -0,0 +1,33 @@
...
+
+class SerdeRegistry[T] extends AbstractDeserializer[T] {
+  private val registry = mutable.Map[Byte, KafkaSerde[_ <: T]]()
+
+  def add[X <: T](version: Byte, serde: KafkaSerde[X]): this.type = {
+    registry += (version -> serde)
+    this
+  }
+
+  def add[X <: T](version: Byte)(implicit codec: GenCodec[T]): this.type = {
```
ghik: Ugh, that generic `X` seems redundant. I overlooked it.

Passarinho4: yeap, sure :)

commons-kafka/src/main/scala/com/avsystem/commons/kafka/VersionedSerializer.scala
Passarinho4: I decided to rewrite KafkaSerde and now it works without copying the array. What do you think?

ghik: Cool, although I agree with @mkej comments about this implementation.

commons-kafka/src/main/scala/com/avsystem/commons/kafka/EventOps.scala
```diff
@@ -0,0 +1,11 @@
+package com.avsystem.commons
+package kafka
+
+import org.apache.kafka.common.serialization.{Deserializer, Serde, Serializer}
+
+trait EventOps {
+  def serde[T](implicit registry: SerdeRegistry[T], ser: Serializer[T]):Serde[T] = new AbstractSerde[T] {
```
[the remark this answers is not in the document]
Passarinho4: Yeap, I need it. When you create KafkaStream, you need to pass Serde[T]. So, you can create object TestEvent extends EventOps{
implicit registry ....;
implicit ser ....;
}
and then you can use TestEvent.serde[TestEvent].

PR #39, opened 2017-09-29
commons-mongo: added Sorting for BsonRef and DocKey

mkej:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/core/ops/BaseSorting.scala
```diff
@@ -0,0 +1,11 @@
+package com.avsystem.commons
+package mongo.core.ops
+
+import com.mongodb.client.model.Sorts
+import org.bson.conversions.Bson
+
+trait BaseSorting extends Any with KeyHandling {
+  def ascending(): Bson = Sorts.ascending(key)
```
ghik: Are these parens needed?

mkej: Nope, not needed at all. Added them for consistency with other no-arg helper methods, will be nicer to use without parens though.

PR #42, opened 2017-10-06
JS size optimization

Starzu:


commons-shared/src/main/scala/com/avsystem/commons/rpc/RPCTypeClasses.scala
```diff
@@ -8,13 +8,8 @@ package rpc
...
-  implicit val asRealRPC: F#AsRealRPC[T] = fullRPCMetadata.asRealRPC
-  implicit val rpcMetadata: F#RPCMetadata[T] = fullRPCMetadata.metadata
-}
-
-case class AllRPCTypeClasses[F <: RPCFramework with Singleton, T](asRealRPC: F#AsRealRPC[T], asRawRPC: F#AsRawRPC[T], metadata: F#RPCMetadata[T])
-object AllRPCTypeClasses {
-  implicit def materialize[F <: RPCFramework with Singleton, T]: AllRPCTypeClasses[F, T] = macro macros.rpc.RPCMacros.allRpcTypeClassesImpl[F, T]
-}
+abstract class RPCTypeClasses[F <: RPCFramework with Singleton, T] {
+  implicit def asRawRPC: F#AsRawRPC[T] = macro macros.rpc.RPCMacros.rpcTypeClassesImplAsRaw[F, T]
```
ghik: How's that supposed to reduce code size if code now these instances get generated at _every_ callsite?
This is completely against what `RPCTypeClasses` was supposed to do - confine macro generated code to companion objects. This is not only about generated code size but also compilation speed and incremental compilation sanity.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -258,30 +265,66 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
...
-          protected def typeRepr = ${tpe.toString}
-          def nullable = ${typeOf[Null] <:< tpe}
-          def readObject(input: $SerializationPkg.ObjectInput, ..$oooFieldsParam): $tpe = {
-            ..${params.map(varDeclaration)}
-            while(input.hasNext) {
-              val fi = input.nextField()
-              fi.fieldName match {
-                case ..${params.map(fieldReadCase)}
-                case _ => fi.skip()
+      if (!isScalaJs) {
```
ghik: Please reduce code duplication and extract the parts that differ between JS and JVM in more finegrained manner.

commons-macros/src/main/scala/com/avsystem/commons/macros/MacroCommons.scala
```diff
@@ -39,6 +39,13 @@ trait MacroCommons {
     Iterator.iterate(enclosingSym)(_.owner).takeWhile(_ != NoSymbol).toList
   }
 
+  final val isScalaJs: Boolean = try {
+    c.typecheck(q"import scalajs.js.native")
```
ghik: Are you sure that `scalajs-library` is _never_ on the classpath for JVM compilation?

Also, I think simpler implementation would be:
```
final val isScalaJs = rootMirror.RootClass.toType.member(TermName("scalajs")) != NoSymbol
```

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -231,7 +232,13 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
       def readField(param: ApplyParam) = {
         val fieldName = nameBySym(param.sym)
         val defaultValueTree = param.defaultValue.orElse(q"fieldMissing($fieldName)")
-        q"${optName(param)}.getOrElse($defaultValueTree)"
+        if (!isScalaJs) q"${optName(param)}.getOrElse($defaultValueTree)"
+        else {
+          q"""
+            if (elements.containsKey(${optName(param).toString})) elements.get(${optName(param).toString}).asInstanceOf[${param.valueType}]
```
ghik: Why not just use `getOrElse`? Are you trying to avoid by-name params? Also, it's weird that `optName` is used here - I would suppose it would be more natural to simply use `fieldName`.
Also, if you're using a map here then for consistency it should also be used in `OutOfOrderFields`

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -258,30 +265,66 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
...
+        q"""
+          new $baseClass(
+            ${tpe.toString},
+            ${typeOf[Null] <:< tpe}
+          ) {
+            ..${params.filterNot(ap => oooFields.contains(nameBySym(ap.sym))).map(depDeclaration)}
+            ..${generated.collect({ case (sym, depTpe) if !oooFields.contains(nameBySym(sym)) => generatedDepDeclaration(sym, depTpe) })}
+            $emptyOooFieldsDecl
+            def readObject(input: $SerializationPkg.ObjectInput, ..$oooFieldsParam): $tpe = {
+              val elements: _root_.java.util.HashMap[String, Any] = new _root_.java.util.HashMap
```
ghik: If that's for JS only, wouldn't you want to use `js.Dictionary` so that it's more lightweight?

commons-shared/src/main/scala/com/avsystem/commons/rpc/RPCTypeClasses.scala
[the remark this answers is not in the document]
Starzu: Right. Anyway we need to find a way to remove dead code from JS.

commons-shared/src/main/scala/com/avsystem/commons/collection/CollectionAliases.scala
```diff
@@ -14,103 +14,103 @@ import scala.{collection => sc}
   */
 trait CollectionAliases {
   type BTraversable[+A] = sc.Traversable[A]
-  final val BTraversable = sc.Traversable
+  final lazy val BTraversable = sc.Traversable
```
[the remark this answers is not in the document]
Starzu: That's right, `lazy val` generates much more code than `val`. On the other hand `lazy val` can be removed in DCE phase. These aliases (mixed in `commons` package object) force compiler to include unused parts of scala collections into JS code. 
We can create different versions of `commons` package object for JVM and JS, but I'm afraid IntelliJ cannot understand it.

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMacros.scala
```diff
@@ -122,24 +121,32 @@ class RPCMacros(ctx: blackbox.Context) extends AbstractMacroCommons(ctx) {
...
-        val badMethodError = s"${member.method} cannot be handled by raw ${variant.rawMethod}"
-        val onFailure = q"throw new Exception($badMethodError)"
         val realInvocation = q"$implName.${member.method}(...$methodArgs)"
-        cq"""
-        (${member.rpcNameString}, $matchedArgs) =>
-          $FrameworkObj.tryToRaw[${member.returnType},${variant.returnType}]($realInvocation,$onFailure)
-        """
+        val expectedHandlerType = getType(tq"$RealInvocationHandlerCls[${member.returnType},${variant.returnType}]")
+
+        c.inferImplicitValue(expectedHandlerType) match {
```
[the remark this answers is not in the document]
ghik: I think the reason for avoiding `inferImplicitValue` was handling of recursively defined types.
Please write a test to make sure this works with recursively-defined RPC interfaces, i.e. there are some getters in this interface that directly or indirectly return the same interface type.

commons-core/src/test/scala/com/avsystem/commons/serialization/JGenCodecTest.scala
```diff
@@ -0,0 +1,39 @@
...
+package serialization
+
+import java.lang.annotation.RetentionPolicy
+
+import com.github.ghik.silencer.silent
+
+import scala.collection.immutable.ListMap
+
+@silent
+class JGenCodecTest extends JCodecTestBase {
```
ghik: Why can't this be also moved to shared sources?

commons-shared/src/main/scala/com/avsystem/commons/collection/CollectionAliases.scala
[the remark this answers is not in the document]
ghik: I see.

I'd prefer to refactor `commons` package object and remove some stuff from it. It will cause incompatibilities, but I still think it's less ugly solution. Perhaps it was a mistake to put all this stuff into the `commons` package by default.

As you said, `val`s can't be DCEd by default because they may have hidden side effects.
I wonder if ScalaJS guys noticed this problem and maybe they're planning to do something about it.

commons-core/src/test/scala/com/avsystem/commons/serialization/JGenCodecTest.scala
Starzu: Missing implementations in Scala.js. I'll move working parts "java collection test" to cross-compiled tests.

commons-shared/src/main/scala/com/avsystem/commons/collection/CollectionAliases.scala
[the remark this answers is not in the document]
ghik: Hmm... maybe we could simply change them into `final def`s?

```scala
final def BTraversable: sc.Traversable.type = sc.Traversable
```

This makes them no longer _stable identifiers_ (e.g. you can't do things like `BTraversable.type`) but I can't think of any use case where it might be needed.

However, this solution does not apply to `val`s in `GenCodec` because there I actually want to ensure single instance.

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMacros.scala
```diff
@@ -122,24 +122,32 @@ class RPCMacros(ctx: blackbox.Context) extends AbstractMacroCommons(ctx) {
...
+        }
       }
 
       def defaultCase(variant: Variant): CaseDef =
         cq"_ => fail(${rpcTpe.toString}, ${variant.rawMethod.name.toString}, methodName, args)"
 
       def methodMatch(variant: Variant, methods: Iterable[ProxyableMember]) = {
-        Match(q"(methodName, args)", (methods.map(m => methodCase(variant, m)) ++ Iterator(defaultCase(variant))).toList)
+        Match(q"(methodName, args)",
+          (methods.map(m => methodCase(variant, m)) ++ Iterator(Some(defaultCase(variant))))
```
ghik: Use `flatMap` instead of `map` and get rid of `Some` and `collect`

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -87,15 +86,15 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
       val depNames = depNamesMap(generated.map(_._1))
 
       def depDeclaration(sym: Symbol, depTpe: Type) =
-        q"lazy val ${depNames(sym)} = ${dependency(depTpe, tcTpe, s"for generated member ${sym.name}")}"
+        if (isScalaJs) q"def ${depNames(sym)} = ${dependency(depTpe, tcTpe, s"for generated member ${sym.name}")}"
```
ghik: I wonder how would that work with DCE if we made a "manual", unsynchronized lazy vals, e.g.

```scala
def depDeclaration(sym: Symbol, depTpe: Type) = {
  val depName = depNames(sym)
  val depVarName = TermName("_" + depName.toString)
  q"""
    private[this] var $depVarName: $depTpe = null
    def $depName = {
      if($depVarName == null) {
        $depVarName = ${dependency(depTpe, tcTpe, s"for generated member ${sym.name}")}
      }
      $depVarName
    }
  """
}
```

This version could be used by both JVM and JS.

build.sbt
```diff
@@ -168,14 +168,14 @@ lazy val `commons-shared` = crossProject.crossType(CrossType.Pure)
       val githubDir = "https://raw.githubusercontent.com/AVSystem/scala-commons"
       s"-P:scalajs:mapSourceURI:$localDir->$githubDir/v${version.value}/"
     },
-    test := {},
+    jsEnv := new org.scalajs.jsenv.nodejs.NodeJSEnv(),
```
ghik: 👍

commons-shared/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
```diff
@@ -321,24 +349,24 @@ object GenCodec extends FallbackMapCodecs with TupleGenCodecs {
...
-  implicit val JByteCodec: GenCodec[JByte] = createNullSafe(_.readByte(), _ writeByte _, allowNull = true)
-  implicit val JShortCodec: GenCodec[JShort] = createNullSafe(_.readShort(), _ writeShort _, allowNull = true)
-  implicit val JIntegerCodec: GenCodec[JInteger] = createNullSafe(_.readInt(), _ writeInt _, allowNull = true)
-  implicit val JLongCodec: GenCodec[JLong] = createNullSafe(_.readLong(), _ writeLong _, allowNull = true)
-  implicit val JFloatCodec: GenCodec[JFloat] = createNullSafe(_.readFloat(), _ writeFloat _, allowNull = true)
-  implicit val JDoubleCodec: GenCodec[JDouble] = createNullSafe(_.readDouble(), _ writeDouble _, allowNull = true)
-  implicit val JBigIntegerCodec: GenCodec[JBigInteger] = createNullSafe(i => new JBigInteger(i.readString()), (o, v) => o.writeString(v.toString), allowNull = true)
-  implicit val JBigDecimalCodec: GenCodec[JBigDecimal] = createNullSafe(i => new JBigDecimal(i.readString()), (o, v) => o.writeString(v.toString), allowNull = true)
+  implicit lazy val BigIntCodec: GenCodec[BigInt] = createNullSafe(i => BigInt(i.readString()), (o, v) => o.writeString(v.toString), allowNull = true)
```
ghik: Can you check the impact of this? I'd rather leave them as they are and I would imagine this is not as impactful as `val`s in `CollectionAliases`.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -258,30 +253,85 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
...
+                  case _ => fi.skip()
+                }
+              }
+              ${applier(params.map(p => p.asArgument(readField(p))))}
+            }
+          """
+        } else {
+          q"""
+            def readObject(input: $SerializationPkg.ObjectInput, ..$oooFieldsParam): $tpe = {
+              val elements: _root_.scala.scalajs.js.Dynamic = _root_.scala.scalajs.js.Object().asInstanceOf[_root_.scala.scalajs.js.Dynamic]
```
ghik: Please make some aliases for these classes or packages, e.g.

```scala
val JsPackage = q"_root.scala.scalajs.js"
```

ghik: I wonder how using a map or dictionary affects performance. I'll try to do some benchmarks in JVM - is there any blessed way of doing benchmarks for JS?
If the map/dictionary approach happens to be not slower for JVM then we could get rid of duplication and use it everywhere.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -258,30 +253,85 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
...
+              }
+              ${applier(params.map(p => p.asArgument(readField(p))))}
+            }
+          """
+        }
+
       q"""
-        new $baseClass with $GenCodecObj.ErrorReportingCodec[$tpe] {
+        new $baseClass(
+          ${tpe.toString},
```
ghik: To avoid unnecessary fields, I would prefer if these were not passed as arguments but as protected method implementations. My intuition is that this should not be a problem with regard to generated code size. But of course correct me if I'm wrong.

ghik:
Ugh, I added a lot of inline comments and github doesn't show them here in the "Conversation" tab. I should've used proper review.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
Starzu: Implementation with HashMap was about 2 times slower.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
Starzu: It generates similar amount of JS code. Simple defs reduce it significantly.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
ghik: I meant to compare the original variable based solution with map based solution.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
Starzu: Method implementation generates code for each created anonymous class.

commons-shared/src/main/scala/com/avsystem/commons/collection/CollectionAliases.scala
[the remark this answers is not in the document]
Starzu: It's even better, if we don't need stable identifiers.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
```diff
@@ -258,30 +253,85 @@ class GenCodecMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) with
...
+            q"""
+              if (!$JsPackage.isUndefined(elements.selectDynamic($fieldName))) {
+                elements.selectDynamic($fieldName).asInstanceOf[${param.valueType}]
+              } else ($defaultValueTree).asInstanceOf[${param.valueType}]
+            """
+          else q"()"
+        }
+      }
+
+      def generateReadObject =
```
ghik: There is still a lot of avoidable duplication in this code.

commons-shared/src/main/scala/com/avsystem/commons/collection/CollectionAliases.scala
```diff
@@ -12,105 +12,106 @@ import scala.{collection => sc}
   * `I` stands for immutable
   * `M` stands for mutable
   */
+//noinspection TypeAnnotation
```
ghik: Please don't do this. Reconfigure IntelliJ and commit settings in `.idea`. Or just add these annotations.

commons-shared/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
ghik: So, is it really worth it? This seems to me like a never-ending story - are we going to keep changing all the `val`s to `lazy val`s just for the sake of JS?

I'm really unhappy about the fact that these optimizations are polluting the codebase and disallowing usage of such basic and simple language elements like `val`s.

commons-core/src/test/scala/com/avsystem/commons/serialization/JGenCodecTest.scala
```diff
@@ -0,0 +1,25 @@
...
+  test("java collection test (TreeMap)") {
+    testWriteReadAndAutoWriteRead[JHashMap[Double, Int]](jDoubleHashMap,
```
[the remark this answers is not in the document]
Starzu: This test checks the order of map entries in the list representation. In JS it is different. ;)

commons-macros/src/main/scala/com/avsystem/commons/macros/MacroCommons.scala
Starzu: Your version doesn't work. It returns `false` for JVM and JS. 

`scalajs-library` contains only `.sjsir` files.

commons-macros/src/main/scala/com/avsystem/commons/macros/TestMacros.scala
```diff
@@ -3,6 +3,7 @@ package macros
 
 import scala.reflect.macros.blackbox
 
+private[commons]
 class TestMacros(val c: blackbox.Context) extends TypeClassDerivation {
```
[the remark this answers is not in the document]
Starzu: If cross-project depends on standard project with "compile->compile;test->test", then generated JS file is broken. It doesn't contain tests code.

commons-macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecMacros.scala
Starzu: What about `def`s in both cases? 

Defs:
```
[info] Benchmark                                                       Mode  Cnt        Score       Error  Units
[info] c.a.c.core.GenCodecBenchmarks.testCaseClassCodec               thrpt   20  2936094.002 â–’ 49606.878  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.conversionTest  thrpt   20   169360.706 â–’  2359.861  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.readTest        thrpt   20   316837.049 â–’  3652.074  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.writeTest       thrpt   20   552318.464 â–’  6536.033  ops/s
```

Lazy vals:
```
[info] Benchmark                                                       Mode  Cnt        Score       Error  Units
[info] c.a.c.core.GenCodecBenchmarks.testCaseClassCodec               thrpt   20  3425750.118 â–’ 48293.942  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.conversionTest  thrpt   20   171961.270 â–’  2343.882  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.readTest        thrpt   20   323721.652 â–’  4722.604  ops/s
[info] c.a.c.rpc.akka.serialization.GenCodecBenchmark.writeTest       thrpt   20   490266.148 â–’ 24286.542  ops/s
```

commons-core/src/test/scala/com/avsystem/commons/serialization/JGenCodecTest.scala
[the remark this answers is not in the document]
ghik: Then I guess I wrote it badly and it should be fixed.

ghik:
I'm going to do some more research to see if can squeeze anything more out of RPC macros.

PR #43, opened 2017-10-09
commons-mongo: added array-specific operations to Filtering

mkej:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/core/ops/Updating.scala
```diff
@@ -3,11 +3,14 @@ package mongo.core.ops
 
 import com.avsystem.commons.mongo.BsonRef
 import com.avsystem.commons.serialization.GenCodec
+import com.mongodb.client.model.Updates
 import org.bson.conversions.Bson
 
 object Updating extends LowPrioUpdating {
   implicit def bsonUpdating(bson: Bson): BsonUpdating = new BsonUpdating(bson)
   implicit def bsonRefIterableUpdating[E: GenCodec, C[T] <: Iterable[T]](bsonRef: BsonRef[C[E]]): BsonRefIterableUpdating[E, C] = {
```
ghik: Are you fine with the fact that this will also catch `Map`s? They don't serialize to arrays so this may fail completely silently in runtime (just a query returning empty result without any error or warning).

commons-mongo/src/main/scala/com/avsystem/commons/mongo/core/ops/Filtering.scala
```diff
@@ -2,11 +2,20 @@ package com.avsystem.commons
...
+import com.mongodb.client.model.Filters
 import org.bson.BsonValue
 import org.bson.conversions.Bson
 
-object Filtering {
+object Filtering extends LowPrioFiltering {
   implicit def bsonFiltering(bson: Bson): BsonFiltering = new BsonFiltering(bson)
   implicit def docKeyFiltering[T](docKey: DocKey[T, _ <: BsonValue]): DocKeyFiltering[T] = new DocKeyFiltering(docKey)
-  implicit def bsonRefFiltering[T](bsonRef: BsonRef[T]): BsonRefFiltering[T] = new BsonRefFiltering(bsonRef)
+  implicit def bsonRefIterableFiltering[E: GenCodec, C[T] <: Iterable[T]](bsonRef: BsonRef[C[E]]): BsonRefIterableFiltering[E, C] = {
```
ghik: Are you fine with the fact that this will also catch `Map`s? They don't serialize to arrays so this may fail completely silently in runtime (just a query returning empty result without any error or warning).

commons-mongo/src/main/scala/com/avsystem/commons/mongo/core/ops/Updating.scala
ghik: I guess this has the same problem with `Map`s being caught as `Iterable`s.

mkej:
I'm planning to create a type class defining which types can be queried as mongo arrays. Stay tuned!

ghik:
So, do you want me to merge that?

mkej:
Well... the API will stay roughly the same after implementing mentioned type class. You can merge this version for now - it's better than not having those operators at all.

PR #49, opened 2018-03-22
JsonString serialization improvements

ddworak:
I've ported a large part of Udash test cases for JSON serialization and, as one might predict, we all turned out to be wrong. 
Once I fixed it/figured out it's not all that bad, I've moved to the JS platform to discover the marvels of number representation there.
Now it seems to work fine on both sides.


commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -52,11 +52,18 @@ class JsonStringInput(reader: JsonReader, callback: AfterElement = AfterElementN
...
 
-  private def matchOr[@specialized(Int, Double, Boolean) T: ClassTag](what: String): T = value match {
+  private def matchOr[@specialized(Boolean) T: ClassTag](what: String): T = value match {
     case t: T => t
     case _ => expected(what)
   }
 
+  private def matchNumericString[T](toNumber: String => T): T = value match {
+    case ns: String =>
+      Try(toNumber(ns)).recover { case t => throw new ReadFailure(s"Invalid number format: $ns", t) }.get
```
ghik: `try` and `catch`?

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -248,23 +248,26 @@ final class JsonReader(val json: String) {
...
+      if (isNext('-') || isNext('+')) {
         advance()
-        if (isNext('-') || isNext('+')) {
-          advance()
-          parseDigits()
-        } else throw new ReadFailure("Expected '+' or '-'")
       }
+      if (isNextDigit) {
+        parseDigits()
+      } else throw new ReadFailure(s"Expected a digit, got ${json.charAt(i)}")
```
ghik: This check looks redundant, `parseDigits` already does that

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -248,23 +248,26 @@ final class JsonReader(val json: String) {
...
     }
 
     val str = json.substring(start, i)
-    if (decimal) str.toDouble else str.toInt
+    str
```
ghik: The problem here is that `readString()` will now work even if there was a number.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -0,0 +1,206 @@
...
+    val oi = input.readObject()
+    val resBuilder = IListMap.newBuilder[String, String]
+    while (oi.hasNext) {
+      val fi = oi.nextField()
+      resBuilder += ((fi.fieldName, fi.readRawJson()))
+    }
+    assert(resBuilder.result() == jsons)
+  }
+
+  def roundtrip[T: GenCodec](name: String)(values: Seq[T]): Unit = {
```
ghik: Take an implicit `org.scalactic.source.Position` to this function

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -0,0 +1,206 @@
...
+    assert(java.lang.Double.isNaN(deserialized))
+  }
+
+  test("scientific") {
+    val value = -1.750470182E9
+    val test = value.toString
+    val serialized = Seq("-1.750470182E+9", "-1.750470182E9", test)
+    val deserialized = serialized.map(read[Double])
+
+    deserialized should contain only value
```
ghik: <rant>
Attempts to make Scala or any other language look like English sentences do not amuse me.
I'm a simple man, I just `assert`
</rant>

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
ddworak: Can you propose an alternative? We're omitting so many other checks for performance reasons and the fact, that the usage pattern explicitly specifies the return type.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
ddworak: Agreed to the first, but I don't want to just pass a single value, because I don't care for sequence ser/de here.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
ddworak: <rant>
I don't care for your personal preferences, but to be perfectly honest I just want to have a beatiful error message when it fails. I can insert the dots if you like them :nerd_face: 
</rant>

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
ghik: Keep a flag along raw value or something. We have benchmarks so we can see if it breaks performance.

ghik:
In general, this is not about marvels of number representation. json.org has it all specified as clearly as possible. But apparently, not for my eye...

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
ghik: If I understand your intention well then I think varargs would be more appropriate than `Seq`.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  test("serialize and deserialize nested case classes") {
+    val test: TestCC = TestCC(5, 123L, 432, true, "bla", 'a' :: 'b' :: Nil)
+    val test2: TestCC = TestCC(-35, 1L, 432, true, "blsddf sdg  \"{,}[,]\"a", 'a' :: 'b' :: Nil)
+    val nested: NestedTestCC = NestedTestCC(-123, test, test2)
+    val serialized = write(nested)
+    val deserialized = read[NestedTestCC](serialized)
+
+    deserialized should be(nested)
```
ghik: `<rant>`
This is also what annoys me about this syntax. Are we using infix/postfix or parens? Here we have a weird mix of both.
`</rant>`

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -0,0 +1,206 @@
...
+    deserialized.i2.binary should be(item.i2.binary)
+    deserialized.i2.list should be(item.i2.list)
+    deserialized.i2.set should be(item.i2.set)
+    deserialized.i2.obj should be(item.i2.obj)
+    deserialized.i2.map should be(item.i2.map)
+  }
+
+
+  test("serialize and deserialize huge case classes") {
+    def cc() = TestCC(Random.nextInt(), Random.nextLong(), Random.nextInt(), Random.nextBoolean(), Random.nextString(Random.nextInt(300)), List.fill(Random.nextInt(300))('a'))
```
ghik: Randomized tests are not cool. You can't really rely on them to tell you that something was fixed after it was broken, so you can't do TDD-style bugfixing.
Property based tests also suffer from this although they tend to scan large set of possible values so it's less harmful.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringOutput.scala
```diff
@@ -44,17 +44,17 @@ final class JsonStringOutput(builder: JStringBuilder) extends BaseJsonOutput wit
...
+      writeString(double.toString)
+    else builder.append(double.toString)
   }
-  def writeDouble(double: Double): Unit = builder.append(double.toString)
   def writeBinary(binary: Array[Byte]): Unit = {
     builder.append('"')
     var i = 0
     while (i < binary.length) {
-      builder.append(f"${binary(i)}%02x")
+      builder.append(f"${binary(i) & 0xff}%02x") //JS has signed chars
```
ghik: What was wrong with the previous version?
I don't care about chars in JS. It's JSON, not JS and I'm just representing Scala bytes in hex. Hex is "unsigned" but I don't care about it as long as it deserializes properly.

ddworak: It produces a different string representation which won't even deserialize with current, logical implementation of an Input. 
```scala
val binary: Byte = -2
val string = f"${binary}%02x" //"fffffffe" in SJS and "fe" on JVM
```
I do care about JS compat and you also should if you want this to be used for e.g. Udash RPCs.

ghik: That is a ScalaJS bug for me. The fact that it works like this in JS is not an excuse.
But I'm fine with the workaround. I checked that only on JVM.

ddworak: It is a bug in the sense that it works differently than on the JVM, but it's not a bug if you just read the docs. %02x is "at least two hexes", so with signed chars this makes sense. It would work exactly the same on a machine with signed char in plain old C.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
ddworak: Maybe I'm just too used to how it works. I'll use `shouldBe` instead.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
ddworak: You do understand that this is a port of a test you most probably reviewed before? :smiling_imp:  I'd rather PBT but it did not work very well for us as you can see ;) I'd rather we detect more issues over a larger number of runs and then write specific test cases for them (as I did for scientific notation) than have a beautiful test suite which only makes me sure I can serialize 42 as a Double.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -0,0 +1,241 @@
...
+      deserialized.string shouldBe item.string
+      deserialized.char shouldBe item.char
+      deserialized.boolean shouldBe item.boolean
+      deserialized.byte shouldBe item.byte
+      deserialized.short shouldBe item.short
+      deserialized.int shouldBe item.int
+      deserialized.long shouldBe item.long
+      deserialized.float shouldBe item.float
+      deserialized.double shouldBe item.double
+      deserialized.binary shouldBe item.binary
```
ghik: Does that compare arrays by element values?

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -26,21 +26,15 @@ object JsonStringInput {
...
+  reader.currentValue match {
+    case JsonStringInput.ListMarker | JsonStringInput.ObjectMarker =>
+    case _ => afterElement()
   }
 
-  def jsonType: String = value match {
+  def jsonType: String = reader.currentValue match {
     case JsonStringInput.ListMarker => "list"
     case JsonStringInput.ObjectMarker => "object"
     case _: String => "string"
```
ghik: This is now wrong.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -68,15 +62,19 @@ class JsonStringInput(reader: JsonReader, callback: AfterElement = AfterElementN
...
     case null => InputType.Null
     case _ => InputType.Simple
   }
 
-  def readNull(): Null = if (value == null) null else expected("null")
-  def readString(): String = matchOr[String]("string")
+  def readNull(): Null = if (reader.currentValue == null) null else expected("null")
+  def readString(): String = {
+    val expected = "string"
+    if (reader.isValueNumeric) throw new ReadFailure(s"Expected $expected but got numeric string: ${reader.currentValue}")
```
ghik: I would say that's just "number", not a "numeric string". In JSON this is just a number, the fact that we keep it as string is our implementation detail.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -50,32 +44,41 @@ class JsonStringInput(reader: JsonReader, callback: AfterElement = AfterElementN
...
+  private def expected(what: String) = throw new ReadFailure(s"Expected $what but got $jsonType: ${reader.currentValue}")
 
-  private def matchOr[@specialized(Int, Double, Boolean) T: ClassTag](what: String): T = value match {
+  private def matchOr[@specialized(Boolean) T: ClassTag](what: String): T = reader.currentValue match {
     case t: T => t
     case _ => expected(what)
   }
 
-  def inputType: InputType = value match {
+  private def matchNumericString[T](toNumber: String => T): T = reader.currentValue match {
```
ghik: This method should check the numeric flag.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -50,32 +44,41 @@ class JsonStringInput(reader: JsonReader, callback: AfterElement = AfterElementN
...
+    case ns: String =>
+      try {
```
ghik: `<pedantic_debracing>`
```
try toNumber(ns) catch
```
`</pedantic_debracing>`

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -26,56 +26,55 @@ object JsonStringInput {
...
   }
 
-  def inputType: InputType = value match {
+  private def matchNumericString[T](toNumber: String => T): T = reader.currentValue match {
+    case ns: String if reader.isNumber || ns == "Infinity" || ns == "-Infinity" || ns == "NaN" =>
+      try toNumber(ns) catch {
+        case e: NumberFormatException => throw new ReadFailure(s"Invalid number format: $ns", e)
+      }
+    case _ =>
+      expected("numeric string")
```
ghik: `expected("number")`

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -26,56 +26,55 @@ object JsonStringInput {
...
+    matchOr[String](expectedTpe)
   }
-  def readDouble(): Double = matchOr[Double]("double number")
-
+  def readBoolean(): Boolean = matchOr[Boolean]("boolean")
+  def readInt(): Int = matchNumericString(_.toInt)
+  def readLong(): Long = matchNumericString(_.toLong)
+  def readDouble(): Double = matchNumericString(_.toDouble)
   def readBinary(): Array[Byte] = {
     val hex = matchOr[String]("hex string")
```
ghik: This one should also check if `!isNumber`, I think we should have a dedicated method instead of using `matchOr[String]`

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -308,20 +296,21 @@ final class JsonReader(val json: String) {
...
-    value = if (i < json.length) json.charAt(i) match {
-      case '"' => parseString()
-      case 't' => pass("true"); true
-      case 'f' => pass("false"); false
-      case 'n' => pass("null"); null
-      case '[' => read(); JsonStringInput.ListMarker
-      case '{' => read(); JsonStringInput.ObjectMarker
-      case '-' => parseNumber()
-      case c if Character.isDigit(c) => parseNumber()
+    val (newValue, newTpe) = if (i < json.length) json.charAt(i) match {
```
ghik: Pairing means boxing.

PR #52, opened 2018-04-03
Relaxed field order preservation requirement for ObjectInput

ghik:
See updated `GenCodec` docs and `ObjectInput` scaladocs for more details.

TLDR: instead of preserving field order, `ObjectInput` may alternatively override `peekField` and provide random field access by field name. This will allow formats using hashtable-based object representation (e.g. native JS representation) to be fully compatible with `GenCodec` - this currently only affects flat sealed hierarchy encoding and possible custom-written codecs.


commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -143,38 +143,43 @@ abstract class FlatSealedHierarchyCodec[T](
...
-            if (defaultCaseIdx != -1) {
-              val defaultCaseName = caseNames(defaultCaseIdx)
-              val wrappedInput = new DefaultCaseObjectInput(fi, input, defaultCaseName)
-              readFlatCase(defaultCaseName, oooFields, wrappedInput, caseDeps(defaultCaseIdx))
-            } else {
-              missingCase(fi.fieldName)
-            }
+    def readCase(caseNameField: FieldInput): T = {
+      val caseName = readCaseName(caseNameField)
+      caseIndexByName(caseName) match {
```
ddworak: 2 unrelated questions about `caseIndexByName`:
1. Why can't we just use `.indexOf(element, index)`?
2. Does it make sense to use a more lookup-friendly data structure?

docs/GenCodec.md
```diff
@@ -64,13 +64,51 @@ bound to any format. It only depends on the fact that this format is capable of
 * `Char`s, `String`s, `Boolean`s and `null`s
```
ddworak: Now that we have both `.nextField` and `.peekField`, we need an example more complex than a single field class and less automatically handled than a case class. I've seen a lot of people struggle with creating 2-field class codec and it's just getting harder.

More generally, should I `peek` or should I `next`? It appears to me that GenCodec won't exactly be completely Input/Output indepedent.

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -143,38 +143,43 @@ abstract class FlatSealedHierarchyCodec[T](
   final def readObject(input: ObjectInput): T = {
     val oooFields = new FieldValues(oooFieldNames, oooDeps, typeRepr)
```
ddworak: Nitpick: are there `aaa` fields as well? :smiley:

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -143,38 +143,43 @@ abstract class FlatSealedHierarchyCodec[T](
...
-              val defaultCaseName = caseNames(defaultCaseIdx)
-              val wrappedInput = new DefaultCaseObjectInput(fi, input, defaultCaseName)
-              readFlatCase(defaultCaseName, oooFields, wrappedInput, caseDeps(defaultCaseIdx))
-            } else {
-              missingCase(fi.fieldName)
-            }
+    def readCase(caseNameField: FieldInput): T = {
+      val caseName = readCaseName(caseNameField)
+      caseIndexByName(caseName) match {
+        case -1 => unknownCase(caseName)
```
ddworak: These barbaric checks could be avoided if `caseIndexByName` would just return `Opt[GenCodec[_: T]`

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -143,38 +143,43 @@ abstract class FlatSealedHierarchyCodec[T](
...
+        case idx => readFlatCase(caseName, oooFields, input, caseDeps(idx))
+      }
+    }
+
+    def read(): T = if (input.hasNext) {
```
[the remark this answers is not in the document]
ghik: At a price of one more level of indentation, but whatever.

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
ghik: 1. `.indexOf(element, index)` would put the underlying `Array` through implicit conversion to `WrappedArray`. I wanted to be as raw as possible.

2. I'm deliberately using `Array` and linear search here (and also in `FieldValues`) because case classes tend to have small number of fields and sealed hierarchies tend to have small number of case classes. Using a map would probably introduce more overhead.

docs/GenCodec.md
ghik: If you want to write a custom codec that creates an object with two-three fields then the way to go is to implement `apply`/`unapply` in companion object or use [fromApplyUnapplyProvider](http://avsystem.github.io/scala-commons/api/com/avsystem/commons/serialization/GenCodec$.html#fromApplyUnapplyProvider[T](applyUnapplyProvider:Any):com.avsystem.commons.serialization.GenCodec[T]). This is now very clearly recommended in the docs.

You should almost never need to implement an object codec manually. If you *really* need to do it then the backend-independent way of doing it would be:

```scala
class Klass(val int: Int, val str: String)
object Klass {
  private def getField(input: ObjectInput, name: String):
    input.peekField(name).orElse(input.nextField().opt).filter(_.fieldName == name)
      .getOrElse(throw new ReadFailure(s"expected field $name")

  implicit val codec: GenCodec[Klass] = GenCodec.createNullableObject(
    input => {
      val int = getField(input, "int").readInt()
      val str = getField(input, "str").readString()
      new Klass(int, str)
    },
    (klass, output) => {
      output.writeField("int").writeInt(klass.int)
      output.writeField("str").writeString(klass.str)
    }
  )
}
```

but really, this should be an absolute last resort.

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
ghik: nooo

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
ghik: Yeah but I can't similarly change `caseIndexByValue` because in `writeObject` I need to access both `caseNames` and `caseDeps` arrays. So I have to work with indices anyway.

docs/GenCodec.md
ghik: `Input` and `Output` interfaces are generally optimized for ease of *implementation* and performance rather than ease of use.

PR #54, opened 2018-04-26
Added missing test cases and fixed bug concerning classes with single vararg parameter

mowczare:
Classes like:
- case class OnlyVarargsCaseClass(strings: String*)
- class OnlyVarargsCaseClassLike(val strings: Seq[String])

couldn't have their GenCodec auto materialized due to the bug in `isCorrectUnapply` method:

Type `singleParam.typeSignature` (T*) was not equal to `elemAdjust(resType)` (Seq[T]).


ghik:
:+1:

PR #56, opened 2018-05-25
Dependency and plugin updates

ddworak:


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.27.3"
+version in ThisBuild := "1.28.0"
```
ghik: Which dependencies are binary incompatible with previous versions?

ddworak: Binary compatibility check succeeds against 1.27.3, but it seems we broke something between 1.27.0 and 1.27.3 undetected. Better late than never? 

We don't seem to run the checks on Travis, maybe we should add it to the pipeline?

ghik: Yes, but I think I didn't care about breakage knowing that 1.27.0-2 have never been actually used before 1.27.3. This is an assumption I allow myself to take because I know that probably nobody in the wild is using `commons`. So I think we should release 1.27.4 and 1.28.0 will contain reworked RPC (soon) which actually breaks compatibility.

ghik:
:+1:

PR #57, opened 2018-05-28
Framework-less, generalized, annotation-based RPC

ghik:
## Generalized RPC/proxy/API manipulation framework.

Example code using stuff described below: [NewRawRpc](https://github.com/AVSystem/scala-commons/blob/uberpc/commons-core/src/test/scala/com/avsystem/commons/rpc/NewRawRpc.scala)

### `AsRaw` and `AsReal`

* `RpcFramework` is no longer required. For backwards compatibility it's kept as a special case of the new framework.
* Two global conversion typeclasses, `AsReal` and `AsRaw` (and a bidirectional conversion, `AsRealRaw`) have been introduced. All real-raw conversions are realized through these typeclasses. This includes both RPC trait conversion and parameter and return value serialization and deserialization.
* There are macros available that materialize `AsReal` and `AsRaw` for real and raw RPC traits. Implicit macros have been removed - all macro materialized instances must be explicitly declared in companion objects of real RPC traits.
* Macros that materialize `AsReal` and `AsRaw` for RPC traits do it solely based on how raw RPC trait looks like. In order to guide and customize the materialization, raw RPC, it's methods and parameters may be annotated with various meta annotations, e.g. `@multi`, `@verbatim`, `@tagged`.
* Raw RPC trait contains raw methods. Every real method must be matched to a single raw method. If a raw method is annotated as `@multi`, it's supposed to handle multiple real methods. This is the most common situation. Whether a real method matches a raw method is determined solely by their signatures.
  * If raw method is not annotated as `@multi` then its name must be the same as real method's name (with respect to `@rpcName`). If raw method is annotated as `@multi` then real method's name doesn't matter. Raw method must take that name as the only parameter in its first parameter list.
  * Return types of raw and real method must match. If raw method is `@verbatim` then return types must be the same. If it's `@encoded` then there must be an appropriate `AsRaw` or `AsReal` conversion between raw and real method's return type. The macro performs an ad-hoc implicit search for that conversion and deems the methods non-matching when it cannot be found. Non-`@multi` raw methods are `@verbatim` by default while `@multi` raw methods are `@encoded` by default.
  * Raw and real parameters must match. Without any meta-annotations put on raw parameters, they must match one-by-one, i.e. have the same types and be in the same order (names don't matter). Raw parameter annotated as `@multi` may match multiple real parameters (spanning multiple parameter lists). Its type must then be a collection: either some subclass of `Iterable[Raw]` or `PartialFunction[String,Raw]` where `Raw` is a type that will be matched against each real parameter type. By default, `@multi` raw parameters are `@encoded` which means that the macro will serialize and deserialize each real parameter value to `Raw` value using `AsRaw` and `AsReal` instances for these types. However - contrary to how method return types are treated - if one of these instance is not available, it will not affect macro's decision about whether raw and real method match. Implicit search errors for parameters are deferred until the very end, after whole macro has been expanded.
  * In other words, raw method's signature serves as a simplistic grammar that real method's signature is parsed with.
* Both methods and parameters may be explicitly filtered using tag annotations. Example: annotations for HTTP methods, e.g. `@GET`, `@POST`, `@PUT`, etc.

The design above removes several limitations:
* There is no longer a single `RawValue` with single serialization strategy for all parameters and return types - each raw method and parameter may use its own type for encoded value with dedicated typeclasses for serialization
* Parameters may be grouped, encoded and collected in various ways instead of just `List[List[RawValue]]`.
* RPC name of real methods must be unique only in scope of a single raw method that they all match to - in previous version RPC name had to be unique in the entire RPC trait

### New RPC metadata

RPC metadata materialization has been reworked in largely similar way as raw-real RPC translation.

* There is no longer any fixed `RPCMetadata` trait. RPC metadata may now be any class. Macro engine reflects on the constructor of that class and based on its parameters determines how it should be materialized. Metadata constructor parameters are annotated in the same way as raw RPC methods and parameters are annotated (i.e. with `@multi`, `@verbatim`, `@encoded`, etc.). This way metadata class mirrors the structure of raw RPC trait. Every raw method is reflected by a constructor parameter in metadata class. That parameter, in turn, holds another metadata class whose constructor parameters mirror raw parameters of that method.
* Apart from the constructor parameters that mirror raw methods or raw parameters, every metadata class may also accept a set additional parameters reified specifically for that particular symbol (real RPC trait, method or parameter). There is another set of annotations that tell the macro engine how to reify these parameters:
  * `@infer` - makes the macro engine search for implicit value of parameter's type. Metadata parameters may simply be declared as `implicit` by themselves and will also be inferred by macro, but with explicit `@infer` annotation you can make non-implicit parameters effectively implicit.
  * `@reifyAnnot` - makes the macro engine reify some annotation(s) from real RPC trait/method/param. By default, macro expects the type of metadata parameter to be a subtype of `StaticAnnotation`. Metadata parameter with `@reifyAnnot` may also be specified as `@optional` or `@multi` to reify optional or repeated annotations.
  * `@hasAnnot[A]` - makes the macro engine reify a `Boolean` flag telling if given real RPC trait/method/param is annotated as `A`.
  * `@reifyName` - makes the macro engine reify a `String` containing RPC trait/method/param name (either with or without respect to `@rpcName`)
  * `@reifyPosition` - makes the macro engine reify an instance of `ParamPosition` which contains detailed information about real parameter's index, index of its parameter list, etc.
  * `@reifyFlags` - makes the macro engine reify an instance of `ParamFlags` which contains a set of flags telling if a real parameter is declared as implicit, by-name, repeated (varargs), has default value or is synthetic (e.g. comes from a context bound).
* Metadata class for real method or real parameter must be a subtype of `TypedMetadata[T]` where `T` must will be matched against real method result type or real parameter type. This way matching of real methods/params to metadata params can mirror matching of real methods/params to raw methods/params.


commons-annotations/src/main/scala/com/avsystem/commons/rpc/RPC.scala
```diff
@@ -1,20 +0,0 @@
...
-  * <ul>
-  * <li>types of arguments must be serializable by uPickle library</li>
-  * <li>return type must be either `Unit`, `Future[T]` where `T` is a type serializable by uPickle library or
-  * another RPC interface</li>
-  * <li>method must not have type parameters</li>
-  * </ul>
-  * RPC interfaces may also have non-abstract members - these will be invoked locally. However, they may invoke
-  * remote members in their implementations.
-  */
-class RPC extends StaticAnnotation
```
Starzu: Can we deprecate it first and remove in the next release?

commons-annotations/src/main/scala/com/avsystem/commons/rpc/RPCName.scala
```diff
@@ -1,9 +0,0 @@
-package com.avsystem.commons
-package rpc
-
-import scala.annotation.StaticAnnotation
-
-/**
-  * You can use this annotation on overloaded RPC methods to give them unique identifiers for RPC serialization.
-  */
-class RPCName(val name: String) extends StaticAnnotation
```
Starzu: Can we deprecate it first and remove in the next release?

commons-core/src/main/scala/com/avsystem/commons/rpc/OptionLike.scala
```diff
@@ -0,0 +1,43 @@
+package com.avsystem.commons
+package rpc
+
+sealed trait OptionLike[O] {
+  type Value
+  def none: O
+  def some(value: Value): O
+  def fold[B](opt: O, ifEmpty: => B)(f: Value => B): B
+}
+trait BaseOptionLike[O, A] extends OptionLike[O] {
```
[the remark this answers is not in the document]
ghik: I use the base trait to do implicit search in macros. This way I avoid using wildcards and the implicit search is simpler. At the same time, I want to avoid the base trait to be extended directly because someone might forget about setting the type member to concrete type. For this reason, `OptionLike` is sealed while `BaseOptionLike` is not.

commons-core/src/main/scala/com/avsystem/commons/rpc/rpc.scala
```diff
@@ -0,0 +1,56 @@
...
+    reusableIdentity.asInstanceOf[AsRealRaw[A, A]]
+
+  def materializeForRpc[R, T]: AsRealRaw[R, T] = macro macros.rpc.RPCMacros.rpcAsRealRaw[R, T]
+}
+
+object RpcMetadata {
+  def materializeForRpc[M[_], T]: M[T] = macro macros.rpc.RPCMacros.rpcMetadata[M[T], T]
+}
+
+trait TypedMetadata[T]
```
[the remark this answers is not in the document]
ghik: Yep, it needs docs.

In short: metadata class for method or parameter must extend `TypedMetadata[X]` where `X` is either the result type of a real method or the type of real parameter. For example:

```scala
class FunctionMetadata[T](...) extends TypedMetadata[Future[T]]
```
and then:
```scala
class RpcMetadata[T](
  functions: Map[String, FunctionMetadata[_]]
  ...
)
```

The macro will then match `Future[_]` against each function actual result type and determine what to substitute for the wildcard.

commons-core/src/test/scala/com/avsystem/commons/rpc/NewRawRpc.scala
```diff
@@ -0,0 +1,147 @@
...
+}
+
+sealed trait untagged extends DummyParamTag
+
+sealed trait RestMethod extends RpcTag
+case class POST() extends RestMethod
+case class GET() extends RestMethod
+case class PUT() extends RestMethod
+
+@methodTag[RestMethod, RestMethod]
```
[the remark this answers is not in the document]
ghik: `RestMethod` (the base trait) as the default means pretty much "no default". This is a test, anyway.

commons-core/src/test/scala/com/avsystem/commons/rpc/NewRawRpc.scala
```diff
@@ -0,0 +1,147 @@
...
+    @multi body: MLinkedHashMap[String, String]): String
+}
+object NewRawRpc extends RawRpcCompanion[NewRawRpc] {
+  override val implicits: this.type = this
+
+  implicit def asRealRawFromGenCodec[T: GenCodec]: AsRealRaw[String, T] = ???
+  implicit def futureAsRealRawFromGenCodec[T: GenCodec]: AsRealRaw[Future[String], Future[T]] = ???
+}
+
+@methodTag[RestMethod, RestMethod]
```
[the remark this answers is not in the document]
ghik: Same here.

commons-annotations/src/main/scala/com/avsystem/commons/rpc/RPCName.scala
ghik: I actually renamed it to `rpcName` :trollface: to retain some naming convention. To be honest, I'd like to have a single annotation for both `GenCodec` (which currently uses `@name`) and RPC. So I will probably introduce some deprecated type aliases (although I'm not sure if they work with annotations).

commons-annotations/src/main/scala/com/avsystem/commons/rpc/RPC.scala
ghik: Hmm, do you have a subclass of `RPC` in Udash? Can we leave that one instead and deprecate it?

commons-annotations/src/main/scala/com/avsystem/commons/rpc/RPCName.scala
Starzu: I just want to get deprecation warning (with migration instructions) instead of class not found error. :)

ghik: I made an alias in [Udash PR](https://github.com/UdashFramework/udash-core/pull/187)

commons-core/src/main/scala/com/avsystem/commons/rpc/AsRealRaw.scala
```diff
@@ -0,0 +1,54 @@
+package com.avsystem.commons
+package rpc
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("don't know how to encode ${T} as ${R}, appropriate AsRaw instance not found")
+trait AsRaw[R, T] {
```
Starzu: Could you replace `R` and `T` with more meaningful names? Maybe `Raw` and `Real`?

commons-annotations/src/main/scala/com/avsystem/commons/rpc/rpcAnnotations.scala
```diff
@@ -0,0 +1,422 @@
...
+/**
+  * Method tagging lets you have more explicit control over which raw methods can match which real methods.
+  * Example:
+  *
+  * {{{
+  * sealed trait RestMethod extends RpcTag
+  * class GET extends RestMethod
+  * class POST extends RestMethod
+  *
+  * @methodTag[RestMethod,GET]
```
ddworak: Maybe default tag is not so useful after all? It seems like a good practice to explicitly write tags anyway (see below), and it's ugly when you don't want to introduce one.

commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -176,6 +180,9 @@ object SharedExtensions extends SharedExtensions {
 
     def ensurePrefix(prefix: String): String =
       if (str.startsWith(prefix)) str else prefix + str
+
+    def indent(ind: String): String =
```
ddworak: Docs, unit tests?

commons-core/src/main/scala/com/avsystem/commons/rpc/AsRawReal.scala
```diff
@@ -0,0 +1,54 @@
...
+@implicitNotFound("don't know how to encode and decode between ${Real} and ${Raw}, appropriate AsRawReal instance not found")
+trait AsRawReal[Raw, Real] extends AsReal[Raw, Real] with AsRaw[Raw, Real]
+object AsRawReal {
+  def create[Raw, Real](asRealFun: Raw => Real, asRawFun: Real => Raw): AsRawReal[Raw, Real] =
+    new AsRawReal[Raw, Real] {
+      def asRaw(real: Real): Raw = asRawFun(real)
+      def asReal(raw: Raw): Real = asRealFun(raw)
+    }
+
+  private val reusableIdentity = new AsRawReal[Any, Any] {
```
ddworak: Premature optimization?

commons-core/src/main/scala/com/avsystem/commons/rpc/RpcUtils.scala
```diff
@@ -0,0 +1,20 @@
+package com.avsystem.commons
+package rpc
+
+import com.avsystem.commons.macros.misc.MiscMacros
+
+/**
+  * @author ghik
+  */
+object RpcUtils {
```
ddworak: These are only used in macro-generated code - maybe we could move this to the macro module?

commons-core/src/main/scala/com/avsystem/commons/rpc/rpcMetadata.scala
```diff
@@ -0,0 +1,72 @@
...
+  * For method metadata, type parameter `T` will be matched against each real method result type.
+  * For parameter metadata, type parameter `T` will be matched against each real parameter type.
+  */
+trait TypedMetadata[T]
+
+/**
+  * Information about real parameter flags and modifiers as defined in Scala code.
+  */
+@transparent
+final case class ParamFlags(rawFlags: Int) extends AnyVal {
```
ddworak: `Enumeration` + `ValueSet` would be more object friendly and less off-by-one error-prone.

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMacros.scala
```diff
@@ -0,0 +1,168 @@
+package com.avsystem.commons
+package macros.rpc
+
+import com.avsystem.commons.macros.AbstractMacroCommons
+
+import scala.reflect.macros.blackbox
+
+abstract class RPCMacroCommons(ctx: blackbox.Context) extends AbstractMacroCommons(ctx) {
```
ddworak: `Rpc`

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMappings.scala
```diff
@@ -0,0 +1,361 @@
+package com.avsystem.commons
+package macros.rpc
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+trait RPCMappings { this: RPCMacroCommons with RPCSymbols =>
```
ddworak: `Rpc`

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMetadatas.scala
```diff
@@ -0,0 +1,379 @@
+package com.avsystem.commons
+package macros.rpc
+
+import scala.annotation.StaticAnnotation
+
+trait RPCMetadatas { this: RPCMacroCommons with RPCSymbols with RPCMappings =>
```
ddworak: `Rpc`

commons-annotations/src/main/scala/com/avsystem/commons/annotation/AnnotationAggregate.scala
```diff
@@ -4,20 +4,47 @@ package annotation
...
+  *   }
   *
   *   case class SomeMongoEntity(@mongoId id: String, data: String)
   * }}}
   *
   * In the above example, applying `@mongoId` annotation on the `id` field has the same effect as if
   * annotations `@name("_id") @outOfOrder` were applied directly on that field.
+  *
+  * NOTE: thanks to the fact that aggregated annotations are applied on a type member
+  * you can pass the arguments of original annotation to aggregated annotations, e.g.
```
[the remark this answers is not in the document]
ghik: Because IntelliJ simply cannot into scaladoc and I stopped caring. Just read it in Sources or in API reference.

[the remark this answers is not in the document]
ghik: IntelliJ cannot into Scaladoc. Among other problems, it removes annotations from code snippets, like in this case. I'm not going to work around it. Just read the sources or API reference.

commons-annotations/src/main/scala/com/avsystem/commons/annotation/AnnotationAggregate.scala
```diff
@@ -4,20 +4,47 @@ package annotation
...
 /**
-  * When an annotation class extends this trait, annotation processing macros (e.g. for `GenCodec` materialization)
-  * will look into annotations of the aggregating annotation itself and apply these annotations as if they were
-  * applied directly on the same target as the aggregating annotation. Example:
+  * Base trait for annotations which aggregate multiple other annotations. This way annotation aggregates
+  * work like "annotation functions" - they are annotations that yield more annotations.
+  *
+  * In order to specify aggregated annotations, the class that extends `AnnotationAggregate` must
+  * redefine the `Implied` dummy type member and apply the aggregated annotations on it. Macro engines
+  * used in `GenCodec` materialization and RPC framework will automatically pick up these annotations.
```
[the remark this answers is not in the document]
ghik: Why? Annotation processing is a part of `MacroCommons` and it should work everywhere.

commons-annotations/src/main/scala/com/avsystem/commons/annotation/AnnotationAggregate.scala
```diff
@@ -4,20 +4,47 @@ package annotation
...
+  *     @rpcName(name) // passing `name` to aggregated annotation
+  *     type Implied
+  *   }
+  * }}}
   */
-trait AnnotationAggregate extends StaticAnnotation
+trait AnnotationAggregate extends StaticAnnotation {
+  /**
+    * Dummy type member meant to be redefined in order to have aggregated annotations applied on it.
+    * These annotations will be automatically picked up by macro engines each time they encounter
```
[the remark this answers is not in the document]
ghik: Definitely not all macro engines in the world.

commons-annotations/src/main/scala/com/avsystem/commons/annotation/AnnotationAggregate.scala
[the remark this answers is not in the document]
ddworak: Non-AVSystem macros.

commons-annotations/src/main/scala/com/avsystem/commons/rpc/rpcAnnotations.scala
```diff
@@ -0,0 +1,422 @@
...
+  * When applied on raw method, it works almost the same way as [[single]] except that it is not required that
+  * matching real method exists in real RPC trait. If there is no matching real method, macro-materialized `AsRaw`
+  * implementation will implement the raw method with a code that throws an exception.
+  *
+  * When applied on raw parameter, specifies that this raw parameter may be matched by a real parameter but this is
+  * not required. Whether a real parameter matches an optional raw parameter is determined by its type and/or tag
+  * (see [[paramTag]] for more information on param tagging).
+  *
+  * Raw parameters marked as [[optional]] must be typed as `Option[T]` (or `Opt[T]`, `OptArg[T]` or whatever type that
+  * has an instance of `OptionLike`). By default, [[optional]] raw parameters are [[verbatim]] which means that the
```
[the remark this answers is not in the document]
ghik: That should be obvious to everyone who knows how implicits work in Scala :sunglasses:

commons-core/src/main/scala/com/avsystem/commons/rpc/AsRawReal.scala
ghik: I'd say a common, fairly harmless practice.

commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
ghik: I'll move it to tests just to avoid writing docs and tests :trollface:

commons-core/src/main/scala/com/avsystem/commons/rpc/rpcMetadata.scala
ghik: Did you just suggest using `scala.Enumeration` to me?

commons-macros/src/main/scala/com/avsystem/commons/macros/rpc/RPCMacros.scala
ghik: wiedziałem

commons-core/src/main/scala/com/avsystem/commons/rpc/rpcMetadata.scala
ddworak: I'd rather `Enumeration` than C-style bit-fiddling. ValueSet is actually quite a nice wrapper around BitSet.

commons-annotations/src/main/scala/com/avsystem/commons/rpc/rpcAnnotations.scala
ghik: If you want all tags to be explicit, you can do it by providing base tag type as the default tag type, which means "no default tag":

```scala
@methodTag[RestMethod,RestMethod]
```

And then every raw method should have `@tagged[Something]` so that every real method must have explicit tag annotation.

commons-core/src/main/scala/com/avsystem/commons/rpc/rpcMetadata.scala
ghik: I refuse to use `Enumeration` which uses reflection and is widely disgusted. I'm going to introduce some safety measures to avoid off-by-one errors in refactorings.

commons-annotations/src/main/scala/com/avsystem/commons/annotation/AnnotationAggregate.scala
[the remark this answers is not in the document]
ghik: I specifically listed `GenCodec` macros and RPC macros - what's the problem?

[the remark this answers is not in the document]
ddworak: Must have misread something :cry:

commons-core/src/main/scala/com/avsystem/commons/rpc/rpcMetadata.scala
Starzu: Conscience clause of Scala developers

PR #58, opened 2018-05-30
Underlying codec in TransparentCodec should be evaluated lazily

ddworak:
It's very easy to forget about initialization order when using `GenCodec.materialize` in conjunction with a handwritten codec.


commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -61,7 +61,7 @@ abstract class ProductCodec[T <: Product](
 abstract class TransparentCodec[T, U](
   protected val typeRepr: String,
   val nullable: Boolean,
-  underlyingCodec: GenCodec[U]
+  underlyingCodec: => GenCodec[U]
```
ghik: Good catch, but I think it should be a `def` implemented with a `lazy val` by macro, like it's done for `dependencies` in `ApplyUnapplyCodec`.

PR #71, opened 2018-06-20
Proposal: provide an implicit Opt(ion) to OptArg conversion

ddworak:
I haven't spent too much time thinking this over, but I've definitely seen a lot of these conversions in `OptArg`-intensive code. Let the Battle for Boilerplatte begin!

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
@@ -11,6 +11,16 @@ object OptArg {
     */
   implicit def argToOptArg[A](value: A): OptArg[A] = OptArg(value)
 
+  /**
+    * This implicit conversion allows you to pass an `Opt` where `OptArg` is required.
+    */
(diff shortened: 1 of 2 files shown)


ghik:
`OptArg` should be used in situations where it is known statically in most callsites whether it will be empty or full. If you find yourself converting `Opt`s and `Option`s to `OptArg` then maybe your API should not use `OptArg` after all?

Plus, if I had to introduce some glue for passing `Opt`s and `Option`s where `OptArg` is expected then it would be a `.toOptArg` converter instead of implicit conversion.

ddworak:
Well, if you have a component which has an inner component and they both take `OptArg`s, you often end up in a situation where you map over an optional argument to provide another optional argument to the inner component. `OptArg`s do not `map` very well (they only `fold`), so you end up going `.toOpt` and again to `.toOptArg`.

ghik:
Then my (initial) opinion is that these components should not use `OptArg`. Maybe that will lead us to conclusion that `OptArg` is not really that useful? Who knows.

PR #79, opened 2018-07-26
Opt#zip

ddworak:
Source and binary incompatible: previously this would resolve to scala.collection.IterableLike#zip and return `Iterable[(A,B)]`

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/misc/Opt.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/misc/Opt.scala
@@ -140,6 +140,9 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with Seria
   @inline def toLeft[X](right: => X): Either[A, X] =
     if (isEmpty) Right(right) else Left(value)
 
+  @inline def zip[B](that: Opt[B]): Opt[(A, B)] =
+    flatMap(l => that.map((l, _)))
+
   /**
     * Apply side effect only if Opt is empty. It's a bit like foreach for Opt.Empty
     *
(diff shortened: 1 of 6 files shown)


ghik:
I thinks that's "only" source incompatible, I don't see a reason why would that be binary incompatible.

But anyway, for now I'm against this because it creates inconsistency between `Opt` and `Option` - I don't think we can reliably add similar `zip` behaviour to `Option`.

ddworak:
Agreed.

This is changed in Scala 2.13. I think we can be in the future already https://github.com/scala/scala/blob/2.13.x/src/library/scala/Option.scala#L321

ghik:
Yes, but the problem is that for `Option` we can't be in the future yet. So we'll end up in situation where `Opt` is in the future but `Option` is still in the not-yet-future which is bad.

ghik:
I'd like to have a build for Scala 2.13 (M4+) but not all dependencies are ready (e.g. Akka).

commons-core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -140,6 +140,9 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with Seria
   @inline def toLeft[X](right: => X): Either[A, X] =
     if (isEmpty) Right(right) else Left(value)
 
+  @inline def zip[B](that: Opt[B]): Opt[(A, B)] =
```
ghik: For consistency, the same should be done for `NOpt` and `OptRef`

ghik: IMHO, creating two lambdas for such a simple operation is a bit overkill. Can we do it in more "manually", e.g. similarly to how Scala 2.13 `Option` is doing it?

commons-core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -140,6 +140,9 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with Seria
...
+    flatMap(l => that.map((l, _)))
```
ghik: IMHO, creating two lambdas for such a simple operation is a bit overkill. Can we do it in more "manually", e.g. similarly to how Scala 2.13 `Option` is doing it?

ddworak: Your level is over 9000 :rofl: Done.

PR #80, opened 2018-07-27
Configurable AsyncContext timeout in JettyRPCFramework handler

ddworak:


commons-jetty/src/main/scala/com/avsystem/commons/jetty/rpc/JettyRPCFramework.scala
```diff
@@ -134,8 +134,8 @@ object JettyRPCFramework extends StandardRPCFramework with LazyLogging {
       invoke(call)(_.fire)
   }
 
-  def newHandler[T](impl: T)(implicit ec: ExecutionContext, asRawRPC: AsRawRPC[T]): Handler =
-    new RPCHandler(asRawRPC.asRaw(impl))
+  def newHandler[T](impl: T, contextTimeoutMillis: Long = 30000)(implicit ec: ExecutionContext, asRawRPC: AsRawRPC[T]): Handler =
```
ghik: `FiniteDuration`?

commons-jetty/src/main/scala/com/avsystem/commons/jetty/rpc/JettyRPCFramework.scala
```diff
@@ -134,8 +137,9 @@ object JettyRPCFramework extends StandardRPCFramework with LazyLogging {
       invoke(call)(_.fire)
   }
 
-  def newHandler[T](impl: T)(implicit ec: ExecutionContext, asRawRPC: AsRawRPC[T]): Handler =
-    new RPCHandler(asRawRPC.asRaw(impl))
+  def newHandler[T](impl: T, contextTimeout: FiniteDuration = FiniteDuration(30000, TimeUnit.MILLISECONDS))(
```
ghik: Why not just `30.seconds`? Looks more readable.

ddworak: To match the description in javax.servlet.AsyncContext#setTimeout

ghik: And why do you want to match it? We're not exposing this method in any way.

ddworak: We were replicating their timeout (unknowingly but still). I couldn't care less really, let's do 0.5 minute or 1/60 hours if you like :rofl:

ghik: Yes, but I just don't see the reason why do we have to express simple 30 seconds as 30000 milliseconds in our API. `AsyncContext` does it because it uses `long` as timeout type. But we're using `FiniteDuration` so why not leverage it to make things as readable as possible to the end user?

ddworak: It's already changed to seconds. I never said you're not right, I'm just tired of discussing nitpicks like these over a course of 5+ comments. #justdoit

ghik: I didn't say it's wrong either. I simply ask when I don't understand why something is done the way it is.

ghik:
LGTM

PR #87, opened 2018-09-20
OkHttp client utilities

Passarinho4:
Hi, I've added Rest Client implementation for OkHttp client. I think it can be useful for lightweight REST clients and mobile applications (Android). Please take a look :)


build.sbt
```diff
@@ -343,3 +345,15 @@ lazy val `commons-akka` = project
       "com.typesafe.akka" %% "akka-testkit" % akkaVersion % Test,
     ),
   )
+
+lazy val `commons-okhttp` = project
+  .dependsOn(`commons-core` % CompileAndTest)
+  .dependsOn(`commons-jetty` % CompileAndTest)
```
Passarinho4: I'm not sure that it works exactly as I expect. I would like to add commons-jetty dependency only for tests.

ghik: `CompileAndTest` is used so that test code of one module can depend on test code of other module. By default the dependency is only established on production code.
If you want a standard test-only dependency, then `% Test` is what you want.

However, I'm a little afraid that we're going to have too many micro modules in `commons` with a separate one created for every dependency. I'm thinking if it would be better to keep one module for various external library integrations and make its external dependencies `Optional`. Currently I have no idea which solution would be better. Both are ugly in some way. WDYT?

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def asHandleRequest(client: OkHttpClient, scheme: String, host: String, port: Int): HandleRequest = {
+    RawRest.safeHandle(request => callback => {
+      val url = new HttpUrl.Builder()
+        .scheme(scheme)
+        .host(host)
+        .port(port)
+
+      request.parameters.path.iterator.foreach(v => url.addPathSegment(v.value))
+      request.parameters.query.iterator.foreach(v => url.addQueryParameter(v._1, v._2.value))
```
ghik: `.foreach { case (name, QueryValue(value)) => url.addQueryParameter(name, value) }`

`.iterator` is redundant.

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import com.avsystem.commons.annotation.explicitGenerics
+import com.avsystem.commons.rest.RawRest.HandleRequest
+import com.avsystem.commons.rest.{HeaderValue, HttpBody, RawRest, RestMetadata, RestResponse}
+import okhttp3.{Call, Callback, HttpUrl, MediaType, OkHttpClient, Request, RequestBody, Response}
+
+object RestClient {
+
+  def apply[@explicitGenerics Real: RawRest.AsRealRpc : RestMetadata]
+  (client: OkHttpClient, scheme: String, host: String, port: Int):Real = RawRest.fromHandleRequest[Real](asHandleRequest(client, scheme, host, port))
```
ghik: Please keep the lines short. My preferred way of breaking method signatures is:

```scala
def method[LongGenerics](
  longParameters: LongTypes
): LongResultType =
  longResult
```

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
```diff
@@ -0,0 +1,65 @@
...
+      val httpReq = new Request.Builder()
+          .url(url.build())
+
+      request.parameters.headers.foreach {
+        case (name, HeaderValue(value)) => httpReq.addHeader(name, value)
+      }
+
+      request.body match {
+        case HttpBody(content, mimeType) =>
+          httpReq.method(request.method.toString, RequestBody.create(MediaType.parse(mimeType), content))
```
ghik: Remember that `mimeType` does not contain `charset` information. Should we add it?

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
```diff
@@ -0,0 +1,65 @@
...
+      client.newCall(httpReq.build())
+        .enqueue(new Callback() {
+          override def onFailure(call: Call, e: IOException): Unit = {
+            callback(Failure(e))
+          }
+          override def onResponse(call: Call, response: Response): Unit = {
+            if(response.isSuccessful) {
+              val body = response.headers()
+                .get("Content-Type")
+                .opt
```
ghik: Does it really deserve a separate line? :smile:

commons-okhttp/src/test/scala/com/avsystem/commons/okhttp/rest/RestTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.scalatest.FunSuite
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.{Seconds, Span}
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class RestTest extends FunSuite with UsesJettyHttpServer with UsesOkHttpClient with ScalaFutures {
```
ghik: Can we merge this test with `AbstractRestCallTest`? I'd like to have a single test suite tested with different client/server implementations.

commons-okhttp/src/test/scala/com/avsystem/commons/okhttp/rest/SomeApi.scala
```diff
@@ -0,0 +1,44 @@
+package com.avsystem.commons
+package okhttp.rest
+
+import com.avsystem.commons.rest.{DefaultRestApiCompanion, GET, Header, POST, Path, Query}
+
+trait SomeApi {
```
ghik: I'd like to have one REST test inferface for all sorts for tests. Currently, that would be `RestTestApi`. Can you adjust it instead of creating more traits which test more or less the same stuff?

commons-okhttp/src/test/scala/com/avsystem/commons/okhttp/rest/UsesJettyHttpServer.scala
```diff
@@ -0,0 +1,27 @@
+package com.avsystem.commons
+package okhttp.rest
+
+import org.eclipse.jetty.server.Server
+import org.scalatest.{BeforeAndAfterAll, Suite}
+
+trait UsesJettyHttpServer extends BeforeAndAfterAll { this: Suite =>
```
ghik: There's already `UsesHttpServer` in `commons-jetty`. Would that be applicable?

commons-okhttp/src/test/scala/com/avsystem/commons/okhttp/rest/UsesOkHttpClient.scala
```diff
@@ -0,0 +1,9 @@
+package com.avsystem.commons
+package okhttp.rest
+
+import okhttp3.OkHttpClient
+import org.scalatest.{BeforeAndAfterAll, Suite}
+
+trait UsesOkHttpClient extends BeforeAndAfterAll { this: Suite =>
```
ghik: If there's no `beforeAll` / `afterAll` setup/destroy logic then I don't see a reason for introducing entire separate trait.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.31.0"
+version in ThisBuild := "1.31.1"
```
ghik: I'll handle the versions myself :wink: I've already done enough incompatible changes that `1.32.0` will be the next one.

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
ghik: Can we just pass `baseUrl` instead of separate scheme, host and port? What if the base URL also needs a base path?

build.sbt
Passarinho4: If we declare external dependencies ```Optional```, it will limit micromodules, but also enforce the user to import them manually. I don't want to change the behavior of other modules, so I decided to create the new one, but ofc your comment is correct. I'm going to merge commons-jetty with commons-okhttp to see the difference and possibly update the PR.

commons-okhttp/src/main/scala/com/avsystem/commons/okhttp/rest/RestClient.scala
Passarinho4: Sure, the URL encoding is handled by the builder.

build.sbt
Passarinho4: Please take a look. I've merged commons-jetty and commons-okhttp into commons-http module. I've fixed parameter encoding, because okHttp doesn't encode + as space in path parameter (Hardcoded plusIsSpace=false parameter in HttpUrl.push method). If you accept these changes, I will correct the documentation.

commons-http/src/main/scala/com/avsystem/commons/http/okhttp/rest/RestClient.scala
```diff
@@ -0,0 +1,79 @@
...
+import com.avsystem.commons.annotation.explicitGenerics
+import com.avsystem.commons.rest.RawRest.HandleRequest
+import com.avsystem.commons.rest.{HeaderValue, HttpBody, PathValue, QueryValue, RawRest, RestMetadata, RestResponse}
+import okhttp3.{Call, Callback, HttpUrl, MediaType, OkHttpClient, Request, RequestBody, Response}
+
+object RestClient {
+
+  def apply[@explicitGenerics Real: RawRest.AsRealRpc : RestMetadata](
+    client: OkHttpClient,
+    scheme: String,
```
ghik: My intention in previous comment was to pass _single_ `baseUrl` which would contain protocol, host, port and path prefix. This would be analogous to Jetty-based `RestClient`. I think it's better because we have just one parameter instead of four and it's easy to build from parts while splitting the already built URL into scheme, host, port and path prefix is way more annoying.

commons-http/src/test/scala/com/avsystem/commons/http/okhttp/rest/OkHttpHttpRestCallTest.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class OkHttpHttpRestCallTest extends AbstractRestCallTest
+  with UsesHttpServer
+  with UsesOkHttpClient {
+
+  override implicit def patienceConfig: PatienceConfig = PatienceConfig(Span(10, Seconds))
+
+  override protected def setupServer(server: Server): Unit = {
```
ghik: Looks like this is copied from `JettyHttpRestCallTest`. Can we share it in base class or something?

build.sbt
ghik: Optional dependencies in case of integration modules like these seem OK to me because there's no way to use e.g. `RestClient` for `OkHttpClient` without creating `OkHttpClient` first. The programmer actually needs to use that dependency himself so it seems OK for me to expect that he will declare it explicitly.

Optional dependency would be very ugly if it was an _internal_ dependency of a `commons-something` module. For example, Akka is an internal dependency of `commons-redis`. It uses Akka, but it (almost) doesn't expose it to the user of Redis driver. So that's a very different situation from what we have in `commons-jetty` etc.

I think we should go all or nothing with merging all integration modules into one. IMHO, that would include `commons-akka`, `commons-jetty`, `commons-spring`, `commons-kafka` and possibly `commons-mongo`. I don't expect you to do it, so you may as well leave `commons-okhttp` for now and some day I'll refactor the project structure.

commons-http/src/main/scala/com/avsystem/commons/http/jetty/rest/RestClient.scala
```diff
@@ -1,5 +1,5 @@
 package com.avsystem.commons
-package jetty.rest
+package http.jetty.rest
```
ghik: I would prefer if we left the packages as they were now. Let's leave project restructuring for later.

commons-http/src/main/scala/com/avsystem/commons/http/okhttp/rest/RestClient.scala
Passarinho4: Yup, I know, but with OkHttp URL builder I can't pass baseUrl which contains protocol, host etc. I need to set all arguments separately. So I decided to expose this API instead of trying to split baseUrl into these parts.  Maybe the most elegant solution requires me to create a special case class for that or use something else, but I want to leave it as simple as possible.

build.sbt
Passarinho4: Ok, I see. If I have enough time I will prepare PR with merging these modules into one. This OkHttp integration can wait. It could be also a great opportunity for me to clean up this commons-kafka module, because it is outdated and some things should be rewritten.

commons-http/src/main/scala/com/avsystem/commons/http/okhttp/rest/RestClient.scala
ghik: I think you can use `HttpUrl.parse(baseUrl).newBuilder` to obtain URL builder set up with base URL.

ddworak:
Maybe we should just integrate with https://github.com/softwaremill/sttp instead for more flexibility? `OkHttp` is one of the supported "wrapped backends" there.

Passarinho4:
@ddworak maybe, but I prepared this integration for my friend who uses OkHttp for Android application. Now I need to test tools like proguard, because the result api.jar size is about 12 mega and this is too much for mobile apps. If it is possible to reduce the size to 1-2 mega, I will update the PR and consider other solutions.  Otherwise, this integration is useless.

ghik:
@Passarinho4 Oh you definitely need something for dead code elimination. REST API trait which uses `DefaultRestApiCompanion` contains a ton of macro generated code which is useless from client side. Scala in general is not really very Android-friendly either - proguard is a must. And I suppose you must use Scala 2.11, because Android doesn't support Java 8 bytecode (or does it now?)

PR #89, opened 2018-10-27
Redis 5.0 support + better blocking operations

ghik:
* Blocking commands (e.g. `BLPOP`) can now be executed by `RedisNodeClient` and `RedisClusterClient` - previously they could only be executed on `RedisConnectionClient`.
  In order to support that, every `RedisNodeClient` now maintains a [separate connection pool](https://github.com/AVSystem/scala-commons/blob/redis5/commons-redis/src/main/scala/com/avsystem/commons/redis/RedisNodeClient.scala#L48) dedicated for blocking commands. Each blocking command reserves the connection until it's finished so that multiple concurrent blocking commands can be executed without waiting on each other to finish. Connections in the blocking pool are allocated as needed and reclaimed after a configurable period of being idle. `NodeConfig` is equipped with new options to control blocking connection pool.
* New sorted set commands: `ZPOPMIN`, `ZPOPMAX`, `BZPOPMIN`, `BZPOPMAX`
* New server commands: `CLIENT ID`, `CLIENT UNBLOCK`, `REPLICAOF` (political correctness)
* `HashKey` type member of `RedisApiSubset` renamed to `Field` so that it can serve both as the type of hash fields and stream entry fields.
* [Streams API](https://github.com/AVSystem/scala-commons/blob/redis5/commons-redis/src/main/scala/com/avsystem/commons/redis/commands/streams.scala) and a [test suite](https://github.com/AVSystem/scala-commons/blob/redis5/commons-redis/src/test/scala/com/avsystem/commons/redis/commands/StreamsApiSuite.scala) which is also a good demonstration of the API.
* Handling cluster `TRYAGAIN` errors by backing out and retrying after a delay.
* New, composable `RetryStrategy` - used for reconnecting, redirection handling and `TRYAGAIN` handling

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
@@ -10,6 +10,7 @@ object OptArg {
     * This implicit conversion allows you to pass unwrapped values where `OptArg` is required.
     */
   implicit def argToOptArg[A](value: A): OptArg[A] = OptArg(value)
+  implicit def intToOptArgLong(int: Int): OptArg[Long] = OptArg(int)
 
   private object EmptyMarker extends Serializable
 
--- a/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
@@ -10,7 +10,10 @@ object OptArg {
     * This implicit conversion allows you to pass unwrapped values where `OptArg` is required.
     */
   implicit def argToOptArg[A](value: A): OptArg[A] = OptArg(value)
+
+  // additional implicits to cover most common, safe numeric promotions
   implicit def intToOptArgLong(int: Int): OptArg[Long] = OptArg(int)
+  implicit def intToOptArgDouble(int: Int): OptArg[Double] = OptArg(int)
 
   private object EmptyMarker extends Serializable
 
--- a/commons-redis/src/main/scala/com/avsystem/commons/redis/RedisApi.scala
+++ b/commons-redis/src/main/scala/com/avsystem/commons/redis/RedisApi.scala
@@ -76,10 +76,10 @@ import com.avsystem.commons.redis.config.ExecutionConfig
   * }}}
   */
 object RedisApi {
-  class Raw[Key: RedisDataCodec, HashKey: RedisDataCodec, Value: RedisDataCodec]
-    extends AbstractRedisApi[Raw, Key, HashKey, Value] with RedisConnectionApi with RedisRawApi {
-    def copy[K, H, V](newKeyCodec: RedisDataCodec[K], newHashKeyCodec: RedisDataCodec[H], newValueCodec: RedisDataCodec[V]) =
-      new Raw[K, H, V]()(newKeyCodec, newHashKeyCodec, newValueCodec)
+  class Raw[Key: RedisDataCodec, Field: RedisDataCodec, Value: RedisDataCodec]
+    extends AbstractRedisApi[Raw, Key, Field, Value] with RedisConnectionApi with RedisRawApi {
+    def copy[K, H, V](newKeyCodec: RedisDataCodec[K], newFieldCodec: RedisDataCodec[H], newValueCodec: RedisDataCodec[V]) =
+      new Raw[K, H, V]()(newKeyCodec, newFieldCodec, newValueCodec)
   }
   /**
     * Entry point for API variants which return [[RawCommand]]s.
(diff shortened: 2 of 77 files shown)


commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
```diff
@@ -10,6 +10,7 @@ object OptArg {
     * This implicit conversion allows you to pass unwrapped values where `OptArg` is required.
     */
   implicit def argToOptArg[A](value: A): OptArg[A] = OptArg(value)
+  implicit def intToOptArgLong(int: Int): OptArg[Long] = OptArg(int)
```
ddworak: You don't seem to use this one. If you want to start adding these, there's a lot more "implicit" conversions to cover even with just numbers.

commons-redis/src/main/scala/com/avsystem/commons/redis/RedisApi.scala
```diff
@@ -228,17 +228,17 @@ object RedisApi {
   }
 }
 
-abstract class AbstractRedisApi[Self[K0, H0, V0] <: AbstractRedisApi[Self, K0, H0, V0], K, H, V](implicit
-  val keyCodec: RedisDataCodec[K], val hashKeyCodec: RedisDataCodec[H], val valueCodec: RedisDataCodec[V])
+abstract class AbstractRedisApi[Self[K0, F0, V0] <: AbstractRedisApi[Self, K0, F0, V0], K, F, V](implicit
+  val keyCodec: RedisDataCodec[K], val fieldCodec: RedisDataCodec[F], val valueCodec: RedisDataCodec[V])
```
ddworak: These don't need to become public.

commons-redis/src/main/scala/com/avsystem/commons/redis/RedisNodeClient.scala
```diff
@@ -78,7 +109,8 @@ final class RedisNodeClient(
   private[redis] def nodeRemoved(): Unit = {
     val cause = new NodeRemovedException(address)
     failure = cause.opt
-    connections.foreach(_ ! Close(cause))
+    connections.foreach(_ ! RedisConnectionActor.Close(cause, stop = true))
```
ddworak: Both `Close` classes seem to always be used with `stop = true`

commons-redis/src/main/scala/com/avsystem/commons/redis/actor/ConnectionPoolActor.scala
```diff
@@ -0,0 +1,79 @@
...
+import akka.actor.{Actor, ActorRef, Props}
+import com.avsystem.commons.redis.NodeAddress
+import com.avsystem.commons.redis.actor.ConnectionPoolActor._
+import com.avsystem.commons.redis.config.{ConnectionConfig, NodeConfig}
+import com.avsystem.commons.redis.exception.RedisException
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.annotation.tailrec
+
+class ConnectionPoolActor(address: NodeAddress, config: NodeConfig, queue: ConcurrentLinkedDeque[QueuedConn])
```
ddworak: Are there multiple actors using the queue? From the actor's perspective it does not look like it has to be concurrent.

commons-redis/src/main/scala/com/avsystem/commons/redis/RedisNodeClient.scala
```diff
@@ -41,9 +42,23 @@ final class RedisNodeClient(
...
-  private val connections = (0 until config.poolSize).iterator.map(createConnection).toArray
+  private val connections = (0 until config.poolSize).iterator.map(newConnection).toArray
   private val index = new AtomicLong(0)
 
+  private val blockingConnectionQueue = new ConcurrentLinkedDeque[QueuedConn]
+  private val blockingConnectionPool =
+    system.actorOf(Props(new ConnectionPoolActor(address, config, blockingConnectionQueue)))
+
+  private def newBlockingConnection(): Future[ActorRef] = {
+    implicit val timeout: Timeout = Timeout(1.second)
```
ddworak: Maybe just inline and pass explicitly?

commons-redis/src/main/scala/com/avsystem/commons/redis/actor/ConnectionPoolActor.scala
```diff
@@ -0,0 +1,79 @@
...
+            queue.addLast(last)
+          }
+        case null =>
+      }
+    }
+    loop(dequeue = false)
+  }
+}
+object ConnectionPoolActor {
+  case class QueuedConn(conn: ActorRef, enqueuedAt: Long)
```
ddworak: final?

commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
ghik: I'm using it in tests. I was hesitant to add this but it became too annoying. Passing `Int`s where `Long` is expected is very common and this promotion isn't caught by `argToOptArg`.

commons-redis/src/main/scala/com/avsystem/commons/redis/RedisNodeClient.scala
ghik: I'd like to leave a possibility to close connection without stopping the actor, even though it's not used now.

commons-redis/src/main/scala/com/avsystem/commons/redis/actor/ConnectionPoolActor.scala
ghik: Queue is used directly in `RedisNodeClient` which is outside actor context.

ddworak: I know that. I just believe that the actor does not have to enforce this requirement in *his* types, he only cares that it's a queue.

commons-core/src/main/scala/com/avsystem/commons/misc/OptArg.scala
ghik: I added `Int => OptArg[Double]` too and IMHO it should be enough. This covers most common and safe numeric promotions.

ddworak: Long -> Double could be an option as well, but I generally agree.

ghik: The fact that scala has `Long => Double` is wrong and don't want to proliferate it.

ghik: OMG there's `Long => Float` :man_facepalming:

PR #92, opened 2018-11-12
Future.sequenceCompleted

ddworak:


commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -327,6 +327,22 @@ object SharedExtensions extends SharedExtensions {
...
+      * @tparam M the type of the `TraversableOnce` of Futures
+      * @param in the `TraversableOnce` of Futures which will be sequenced
+      * @return the `Future` of the `TraversableOnce` of results
+      */
+    def sequenceCompleted[A, M[X] <: TraversableOnce[X]](in: M[Future[A]])(
+      implicit cbf: CanBuildFrom[M[Future[A]], Try[A], M[Try[A]]],
+      cbf2: CanBuildFrom[M[Try[A]], A, M[A]],
+      executor: ExecutionContext
+    ): Future[M[A]] =
+      Future.traverse(in)(_.transformNow(Success(_))).mapNow(tries => Try.sequence(tries).get)
```
ghik: I was thinking about other implementations that would avoid double `CanBuildFrom` and double collection copying, came up with this... *thing*:
```scala
def sequenceCompleted[A, M[X] <: TraversableOnce[X]](in: M[Future[A]])(
  implicit cbf: CanBuildFrom[M[Future[A]], A, M[A]],
  executor: ExecutionContext
): Future[M[A]]
  val counter = new AtomicInteger(1)
  val barrier = Promise[Unit]()
  val result = Future.traverse(in) { fut =>
    counter.incrementAndGet()
    fut.onCompleteNow { case _ => 
      if(counter.decrementAndGet() == 0) {
        barrier.success(())
      }
    }
    barrier.flatMapNow(_ => fut)
  }
  if(counter.decrementAndGet() == 0) {
    barrier.success(())
  }
  result
}
```

You are free to do whatever you want with this. For instance, you may consider it an overkill and ignore it.

However, as a separate remark - why not implement `traverseCompleted` which is more general and then `sequenceCompleted`  could be just a special case of `traverseCompleted`?

ddworak: You inspired me to try another implementation for this, a bit simpler than what you proposed above - please check it out.

`traverseCompleted` will be useful as well, done.

PR #95, opened 2018-11-22
Ability to intercept exception thrown during rest call processing

mjjaniec:
I was looking for place where I could catch REST exceptions and log them. But it seems that they are "safely" consumed and converted to http response without any trace in library-client side. 
I added callback that will be called for any exception with default implementation that simply logs it, so default behavior is sane and it could be easily tweaked

### DIFF
--- a/commons-jetty/src/main/scala/com/avsystem/commons/jetty/rest/RestHandler.scala
+++ b/commons-jetty/src/main/scala/com/avsystem/commons/jetty/rest/RestHandler.scala
@@ -2,7 +2,7 @@ package com.avsystem.commons
 package jetty.rest
 
 import com.avsystem.commons.annotation.explicitGenerics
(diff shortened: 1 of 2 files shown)


ghik:
@mjjaniec I'd suggest you to make a PR to [Udash](https://github.com/UdashFramework/udash-core) which contains the REST framework ported from commons. This will be the actively maintained framework while `commons-rest` exists only for backwards compatibility.

If you want a quick solution for `coiote-dm` without migrating to Udash REST, then I would suggest creating a custom implementation of REST servlet - this should be fairly simple.

mjjaniec:
No, I do not want to migrate to Udash as we have no plans to use it, from the other hand w do not really need commons-jetty.
We just used 2 methods, in the end, I incorporated few lines to our code base and removed commons-jetty dependence entirely.

PR #96, opened 2018-11-27
Expose current BsonType as metadata in BsonInput

ddworak:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
```diff
@@ -3,14 +3,25 @@ package mongo
...
 import org.bson.types.ObjectId
 
 object ObjectIdMarker extends TypeMarker[ObjectId]
 
+object BsonTypeMetadata extends InputMetadata[BsonType]
+
 trait BsonInput extends Any with InputAndSimpleInput {
   def readObjectId(): ObjectId
 
+  protected def bsonType: BsonType
```
ghik: I think it's fine to make it public, like `jsonType` in `JsonStringInput`.

PR #99, opened 2018-12-21
Fix of serialization and parsing of floats to/from JSON + more efficient implementation of indention during JSON serialization

plokhotnyuk:
Currently most floats are stored and parsed as doubles:

https://github.com/AVSystem/scala-commons/blob/fed6f314f62cc817b1b32f95d555d6905d6f4f8b/commons-core/src/main/scala/com/avsystem/commons/serialization/InputOutput.scala#L87

https://github.com/AVSystem/scala-commons/blob/fed6f314f62cc817b1b32f95d555d6905d6f4f8b/commons-core/src/main/scala/com/avsystem/commons/serialization/InputOutput.scala#L229

It leads to wasting space and rounding errors.

This PR aim to fix that for JSON representation.

Also, binary representation should be fixed too, but it can require migration of already stored data or pairs of online counterparties where binary representation was used.

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringOutput.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringOutput.scala
@@ -3,6 +3,8 @@ package serialization.json
 
 import com.avsystem.commons.serialization._
 
+import scala.annotation.tailrec
+
 object JsonStringOutput {
   def write[T: GenCodec](value: T, options: JsonOptions = JsonOptions.Default): String = {
     val sb = new JStringBuilder
@@ -17,8 +19,10 @@ object JsonStringOutput {
 trait BaseJsonOutput {
   protected final def indent(builder: JStringBuilder, indentSize: OptArg[Int], depth: Int): Unit =
     indentSize match {
-      case OptArg(size) => builder.append('\n').append(" " * (depth * size))
       case OptArg.Empty =>
+      case OptArg(size) =>
+        builder.append('\n')
+        writeSpaces(builder, depth * size)
     }
 
   protected final def writeJsonString(builder: JStringBuilder, str: String, ascii: Boolean): Unit = {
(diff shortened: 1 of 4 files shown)


ghik:
@plokhotnyuk Can you elaborate on what's wrong with binary representation?

ghik:
@plokhotnyuk no that's fine, I was just confused a bit.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringOutput.scala
```diff
@@ -17,8 +17,15 @@ object JsonStringOutput {
 trait BaseJsonOutput {
   protected final def indent(builder: JStringBuilder, indentSize: OptArg[Int], depth: Int): Unit =
     indentSize match {
-      case OptArg(size) => builder.append('\n').append(" " * (depth * size))
       case OptArg.Empty =>
+      case OptArg(size) =>
+        builder.append('\n')
+        val spaceNum = depth * size
+        var i = 0
```
ghik: Looks like whitespace appending is implemented twice (here and in `writeField`). Can you extract a common method for that?

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringOutput.scala
```diff
@@ -90,7 +100,8 @@ final class JsonStringOutput(builder: JStringBuilder, options: JsonOptions = Jso
       builder.append('"')
       var i = 0
       while (i < binary.length) {
-        builder.append(f"${binary(i) & 0xff}%02x") //JS has signed chars
+        val b = binary(i)
+        builder.append(toHex((b >> 4) & 15)).append(toHex(b & 15))
```
ghik: Please use hexadecimal notation in bitwise operations, ie. `0x0F` or `0xF` instead of `15`

ghik:
Nice, thanks!

plokhotnyuk:
Will you consider to fix the source of problem in lines mentioned in the PR description and finally switch binary representation of stored floats to 4-byte format?

ghik:
@plokhotnyuk by "binary representation" are you referring to other/potential implementations of `Input`/`Output` (the ones that store binary data)?

`Float` is IMHO rarely used so I'd like to avoid making `readFloat()` and `writeFloat()` abstract methods. Converting to `Double` seems like a sensible default representation. If someone really wants to make it more compact then it's always possible to optimize by overriding `readFloat()` and `writeFloat()`.

Correct me if I'm wrong but the conversion between `Float` and `Double` *itself* does not introduce any information loss, right? It's the conversion to decimal representation which causes problems. That's why it makes sense to handle `Float`s manually in JSON but I don't see it being necessary for binary formats.

plokhotnyuk:
You are right, the round-trip conversion `Float.toDouble.toFloat` is lossless. 

But for binary only parsing/serialization users should override `readFloat()/writeFloat()` to use a compact 4-byte format.

Also, while default implementation that uses `Double` for parsing and serialization `Float` values are in base traits the same problems (extra rounding and redundant digits) can happens in other implementations of `SimpleInput/SimpleOutput` for Redis or for Mongo (when non-binary implementation is used).

All these problems can be omitted by removing those default implementations which uses `Double` for `Float`.

ghik:
Currently I think making `readFloat()` and `writeFloat()` abstract methods is not worth the cost, because:
* it's a painful backwards compatibility breach
* `Float` is IMHO a very rarely used type, it's almost always better to use `Double`
* forgetting to specialize for `Float` only results in slightly more error in decimal representation and potentially a bit more space/memory used than possible. I think these are mild dangers.

PR #100, opened 2019-01-18
SBT 1.2.8

ddworak:
see: https://developer.lightbend.com/blog/2018-12-30-sbt-1-2-8/index.html

* crossScalaVersions should be set to Nil on the root aggregating project
* simplified noPublishSettings
* Scala.Js 0.6.26


ghik:
thx!

PR #101, opened 2019-02-05
opt2Iterable should return an immutable Iterable

ddworak:


ghik:
Closing because `Opt` should be as close in behaviour to `Option` as possible. I have a fear of some unforeseen type inference consequences, especially since we're dealing with implicit conversions.

PR #125, opened 2019-05-28
Support varargs in Delegation

ddworak:


ghik:
Omg, someone actually finally used this.

PR #138, opened 2019-06-25
Codecs now have an ability to declare list or object size explicitly

ghik:
This is especially useful for CBOR which has two representations for lists and objects (definite and indefinite length). Previously, it was impossible to use definite-length representation with standard `Output` API.

Also, `ListInput` and `ObjectInput` may now hint the codec about their size which may be used to optimize `GenCodec` implementation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ val previousCompatibleVersions = Set("1.34.8")
 
 val commonSettings = Seq(
   organization := "com.avsystem.commons",
-  crossScalaVersions := Seq("2.13.0", "2.12.8"),
+  crossScalaVersions := Seq("2.12.8", "2.13.0"),
   scalaVersion := crossScalaVersions.value.head,
   compileOrder := CompileOrder.Mixed,
   scalacOptions ++= Seq(
--- a/commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
@@ -521,7 +521,11 @@ object SharedExtensionsUtils extends SharedExtensions {
       * In Scala 2.13 this extension method is always be hidden by an actual method available on `IterableOnce`.
       */
     def knownSize: Int = coll match {
+      case c: BIterable[_] if c.isEmpty => 0
       case is: BIndexedSeq[_] => is.size
+      case _: IListMap[_, _] => -1
+      case m: BMap[_, _] => m.size
+      case s: BSet[_] => s.size
       case _ => -1
     }
 
--- a/commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
@@ -360,9 +360,7 @@ object GenCodec extends RecursiveAutoCodecs with TupleGenCodecs {
     o.sizePolicy match {
       case SizePolicy.Ignored =>
       case SizePolicy.Optional =>
-        if (coll.isEmpty) {
-          o.declareSize(0)
-        } else coll.knownSize match {
+        coll.knownSize match {
           case -1 =>
           case size => o.declareSize(size)
         }
--- a/commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
@@ -263,7 +263,7 @@ object GenCodec extends RecursiveAutoCodecs with TupleGenCodecs {
     def size(value: T): Int
 
     protected final def declareSizeFor(output: ObjectOutput, value: T): Unit =
-      if (output.sizePolicy != SizePolicy.Never) {
+      if (output.sizePolicy != SizePolicy.Ignored) {
         output.declareSize(size(value))
       }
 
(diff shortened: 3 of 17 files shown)


commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -515,7 +515,11 @@ object SharedExtensionsUtils extends SharedExtensions {
       * In Scala 2.13 this extension method is always be hidden by an actual method available on `IterableOnce`.
       */
     def knownSize: Int = coll match {
+      case c: BIterable[_] if c.isEmpty => 0
       case is: BIndexedSeq[_] => is.size
+      case _: IListMap[_, _] => -1
```
ddworak: Why isn't it 0 for empty?

commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
```diff
@@ -343,14 +356,43 @@ object GenCodec extends RecursiveAutoCodecs with TupleGenCodecs {
...
     nullableSimple(i => UUID.fromString(i.readString()), (o, v) => o.writeString(v.toString))
 
-  private implicit class IterableOnceOps[A](private val coll: IterableOnce[A]) extends AnyVal {
-    def writeToList(lo: ListOutput)(implicit writer: GenCodec[A]): Unit =
-      coll.iterator.foreach(writer.write(lo.writeElement(), _))
+  private def declareSize(o: SequentialOutput, coll: BIterable[_]): Unit =
+    o.sizePolicy match {
+      case SizePolicy.Ignored =>
+      case SizePolicy.Optional =>
+        if (coll.isEmpty) {
```
ddworak: This case should be possible to handle in the second branch. Tests seem to confirm.

commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
```diff
@@ -370,7 +416,15 @@ object GenCodec extends RecursiveAutoCodecs with TupleGenCodecs {
   }
 
   implicit def arrayCodec[T: ClassTag : GenCodec]: GenCodec[Array[T]] =
-    nullableList[Array[T]](_.iterator(read[T]).toArray[T], (lo, arr) => arr.iterator.writeToList(lo))
+    nullableList[Array[T]](_.iterator(read[T]).toArray[T], (lo, arr) => {
+      lo.declareSize(arr.length)
+      @tailrec def loop(idx: Int): Unit =
```
ddworak: Isn't this premature optimization vs `arr.foreach`? I remember the differences in performance were marginal when I last measured it.

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/CborOutput.scala
```diff
@@ -184,30 +173,68 @@ class CborOutput(out: DataOutput, fieldLabels: FieldLabels)
...
-  extends BaseCborOutput(out) with ListOutput {
+abstract class CborSequentialOutput(
+  out: DataOutput,
+  override val sizePolicy: SizePolicy
+) extends BaseCborOutput(out) with SequentialOutput {
+
+  protected[this] var size: Int = -1
+  protected[this] var fresh: Boolean = true
+
+  protected def writeInitial(major: MajorType): Unit =
```
ddworak: `final`?

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
```diff
@@ -101,12 +102,16 @@ abstract class ProductCodec[T <: Product](
...
-    var i = 0
-    while (i < value.productArity) {
-      writeField(output, i, value.productElement(i))
-      i += 1
-    }
+  def size(value: T): Int = value.productArity
+
+  final def writeFields(output: ObjectOutput, value: T): Unit = {
+    val size = value.productArity
+    @tailrec def loop(idx: Int): Unit =
```
ddworak: You could just use `productIterator`

build.sbt
```diff
@@ -46,7 +46,7 @@ val previousCompatibleVersions = Set("1.34.8")
 
 val commonSettings = Seq(
   organization := "com.avsystem.commons",
-  crossScalaVersions := Seq("2.13.0", "2.12.8"),
+  crossScalaVersions := Seq("2.12.8", "2.13.0"),
```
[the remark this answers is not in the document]
ghik: Yeah, mostly because current IntelliJ still has some problems with 2.13 and not all modules are migrated.

commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
ghik: It is, look two lines up.

commons-core/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala
ghik: Well, it's just a loop. It wasn't specifically intended as optimization.

commons-core/src/main/scala/com/avsystem/commons/serialization/macroCodecs.scala
ghik: I can avoid instantiating the iterator. You may consider this a premature optimization but I believe it to be no less readable than `productIterator` and it's _potentially_ faster than it.

PR #140, opened 2019-07-02
Added mapOr (type inference friendly fold) to option-like classes

ghik:


ghik:
@ddworak I'd prefer to have it inlined just like everything else.

PR #141, opened 2019-07-03
flatCollect should not consume an iterator eagerly

ddworak:


commons-core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -553,11 +553,8 @@ object SharedExtensionsUtils extends SharedExtensions {
 
     def findOpt(p: A => Boolean): Opt[A] = it.find(p).toOpt
 
-    def flatCollect[B, That](f: PartialFunction[A, IterableOnce[B]])(implicit bf: BuildFrom[C[A], B, That]): That = {
-      val b = bf.newBuilder(coll)
-      it.foreach(f.runWith(_.iterator.foreach(b += _)))
-      b.result()
-    }
+    def flatCollect[B](f: PartialFunction[A, IterableOnce[B]])(implicit fac: Factory[B, C[B]]): C[B] =
+      coll.iterator.collect(f).flatten.to(fac)
```
ghik: I assume it's the overload for `Iterator` that fixes the problem, is this change necessary then?

ddworak: In 2.13 the overload wouldn't be needed with this implementation.

ghik:
Oops I didn't notice this was targeting `master`

PR #149, opened 2019-08-01
mongo: added support for decimal128

mkej:


commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonGenCodecs.scala
```diff
@@ -16,4 +18,9 @@ object BsonGenCodecs {
   )
   implicit val objectIdKeyCodec: GenKeyCodec[ObjectId] =
     GenKeyCodec.create(new ObjectId(_), _.toHexString)
+
+  implicit val decimal128Codec: GenCodec[Decimal128] = GenCodec.nullable(
```
ghik: Would you like to give it a `GenKeyCodec` too? (`ObjectId` has one, so for the sake of consistency)

PR #159, opened 2019-08-30
Redis: fall back to seed nodes for cluster refresh after given timeout

ghik:


kiemlicz:
Only one thing missing here: test
Is it possible to simulate the restart of the Redis instances with port changed?

ghik:
Manual tests show that this is working.
Designing an automatic test scenario for this in seriously annoying and I don't want it to block this PR. IMO it can be done later.

PR #269, opened 2020-06-16
Track config epoch in Redis driver and ignore old cluster states

ghik:


commons-redis/src/main/scala/com/avsystem/commons/redis/actor/ClusterMonitoringActor.scala
```diff
@@ -49,6 +49,7 @@ final class ClusterMonitoringActor(
   private val connections = new MHashMap[NodeAddress, ActorRef]
   private val clients = new MHashMap[NodeAddress, RedisNodeClient]
   private var state = Opt.empty[ClusterState]
+  private var lastEpoch: Long = 0
```
kiemlicz, line 52: Could you create test that simulates replica with very old epoch that is being queried?

ghik: It's probably doable but does not seem very straightforward tbh...

kiemlicz: If you control the `nodes.conf` then you should have the ability to set epoch there (I'm not 100% sure it will work)
CLUSTER SET-CONFIG-EPOCH won't work (via [doc](https://redis.io/commands/cluster-set-config-epoch))

Thus if not the `nodes.conf` way then LGTM

ghik: Since this is a very non-straightforward test to set up and I'm not sure whether manual hacking of `nodes.conf` won't freak Redis out, I'm not going to do it now. The change is simple and regression testing should be enough to confirm that it's ok.

PR #298, opened 2020-09-04
Fix writing _id field in BsonOutput with 4.0 driver

sebaciv:
In 4.0 driver there is new special implementation of the writer for _id field when doing inserts `com.mongodb.internal.connection.IdHoldingBsonWriter`. It creates internally completely new instance of the writer that ends up in the wrong state when calling `org.bson.BsonWriter.writeStartDocument(java.lang.String)` (and _id is an object!). Resulting exception:

`Caused by: sbt.ForkMain$ForkError: org.bson.BsonInvalidOperationException: A Name value cannot be written to the root level of a BSON document. at org.bson.AbstractBsonWriter.throwInvalidState(AbstractBsonWriter.java:740) at org.bson.AbstractBsonWriter.writeName(AbstractBsonWriter.java:531) at org.bson.AbstractBsonWriter.writeStartDocument(AbstractBsonWriter.java:271) at com.mongodb.internal.connection.BsonWriterDecorator.writeStartDocument(BsonWriterDecorator.java:43) at com.mongodb.internal.connection.LevelCountingBsonWriter.writeStartDocument(LevelCountingBsonWriter.java:36) at com.mongodb.internal.connection.IdHoldingBsonWriter.writeStartDocument(IdHoldingBsonWriter.java:68) at com.avsystem.commons.mongo.BsonWriterNamedOutput.writeObject(BsonWriterOutput.scala:72)`

### DIFF
--- a/commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonWriterOutput.scala
+++ b/commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonWriterOutput.scala
@@ -69,7 +69,8 @@ final class BsonWriterNamedOutput(escapedName: String, bw: BsonWriter, override 
     new BsonWriterListOutput(bw, legacyOptionEncoding)
   }
   override def writeObject(): BsonWriterObjectOutput = {
-    bw.writeStartDocument(escapedName)
+    bw.writeName(escapedName) // org.bson.BsonWriter.writeStartDocument(java.lang.String) fails when writing _id in 4.0 driver
+    bw.writeStartDocument()
     new BsonWriterObjectOutput(bw, legacyOptionEncoding)
   }
   override def writeObjectId(objectId: ObjectId): Unit =


commons-mongo/src/main/scala/com/avsystem/commons/mongo/BsonWriterOutput.scala
```diff
@@ -69,7 +69,8 @@ final class BsonWriterNamedOutput(escapedName: String, bw: BsonWriter, override
     new BsonWriterListOutput(bw, legacyOptionEncoding)
   }
   override def writeObject(): BsonWriterObjectOutput = {
-    bw.writeStartDocument(escapedName)
+    bw.writeName(escapedName) // org.bson.BsonWriter.writeStartDocument(java.lang.String) fails when writing _id in 4.0 driver
```
ghik, line 72: Can you give a quick explanation what happened?

ghik:
Oops sorry @ddworak I was too impatient and didn't notice you were requested as reviewer.

ddworak:
@ghik don't worry about it. I'm happy with the workaround, but I don't think it's safe to use this anyway. We should update mongo driver to 4.x here first, the changes between these versions are huge and we cannot assume binary compatibility.

Also @sebaciv, do you know why `com.mongodb.internal.connection.IdHoldingBsonWriter` implements `writeStartDocument` differently? When should we worry about bypassing this logic?

PR #299, opened 2020-09-04
Mongo driver 4.1.0

sebaciv:
Bump mongo driver to 4.1.0 and remove deprecated usages


commons-mongo/src/main/scala/com/avsystem/commons/mongo/async/GenCodecCollection.scala
```diff
@@ -3,14 +3,15 @@ package mongo.async
 
 import com.avsystem.commons.mongo.core.GenCodecRegistry
 import com.avsystem.commons.serialization.GenCodec
-import com.github.ghik.silencer.silent
-import com.mongodb.async.client.{MongoCollection, MongoDatabase}
+import org.mongodb.scala.{MongoCollection, MongoDatabase}
 
 object GenCodecCollection {
```
ddworak, line 8: Does it differ from `com.avsystem.commons.mongo.scala.GenCodecCollection`? I guess we should remove this one now.

sebaciv: Okay, deleted. I also moved to `MongoObservableExtensions` to `scala` package

ddworak: Let's call it MongoScalaObservableExtensions as they're no longer generic to Mongo (and the Mongo ones have different interface based on reactive streams)

ddworak (approved):

ddworak:
Please fix tests as well ;)

PR #317, opened 2020-10-29
Well typed MongoDB API

ghik:
This PR implements a Scala-idiomatic layer over the Reactive Streams driver for MongoDB.

* `GenCodec` is used for serialization, similarly to already existing `GenCodecCollection` variants.
* Raw `Publisher`s returned by the Reactive Streams driver are translated into Monix `Observable`s and `Task`s.
  * this gives us a very rich API with transformations like `map`, `filter`, etc.
  * it is straightforward to convert between `Observable`s, `Task`s and `Future`s, e.g. we can turn an `Observable[T]` into a `Task[List[T]]` which can then be turned into a `Future[List[T]]`
* Well typed representations of various BSON documents used in database commands: queries, projections, updates, sort orders, indices, etc.
  * Instead of using raw `Bson`, we have `MongoDocumentFilter[E]`, `MongoProjection[E, T]`, `MongoUpdate[E]`, `MongoDocumentOrder[E]`, `MongoIndex[E]` (`E` denotes type of the entity)
* References to document fields are well typed as `MongoPropertyRef[E, T]` (similarly to already existing `BsonRef[E, T]`)
  * just like with `BsonRef`, you can create these references without using raw strings, e.g. `MyEntity.ref(_.myField)`
  * property references directly expose rich API for creating filters, projections, updates, etc.
* Better support for sealed hierarchies
  * for example, it's easy to create a query or projection that limits results to selected case class or intermediate subtrait in a sealed hierarchy

Before reviewing the actual code, one might want to look at the [tutorial](https://github.com/AVSystem/scala-commons/blob/mongo-api/docs/TypedMongo.md).
Good entry point for review is [TypedMongoCollection](https://github.com/AVSystem/scala-commons/pull/317/files#diff-16757a8bf8bf944af4b9d4676b88f0f340b4fc40492a9d5e55c6016a47e1e87b)


commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoDocumentFilter.scala
```diff
@@ -0,0 +1,137 @@
...
+          val path = fullPath(prop.propertyPathString)
+          findFilterDoc(path).put(path, opFilter.toOperatorsBson)
+      }
+    }
+  }
+}
+object MongoDocumentFilter {
+  private[this] val reusableEmpty = Empty()
+
+  def empty[E]: MongoDocumentFilter[E] = reusableEmpty.asInstanceOf[MongoDocumentFilter[E]]
```
ddworak, line 117: A filter that's never satisfied is often useful as well (e.g. fold + or). I think `Filters.exists("_id", false)` is always correctly recognized as useless by Mongo when executing such queries, doesn't break index use etc.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoIndex.scala
```diff
@@ -0,0 +1,50 @@
+package com.avsystem.commons
+package mongo.typed
+
+import com.avsystem.commons.misc.{AbstractValueEnum, AbstractValueEnumCompanion, EnumCtx}
+import org.bson.{BsonDocument, BsonValue}
+
+
+case class MongoIndex[E](fields: Vector[(MongoPropertyRef[E, _], MongoIndexType)]) {
```
ddworak, line 8: What about index properties? https://docs.mongodb.com/manual/core/index-properties/

commons-mongo/src/main/scala/com/avsystem/commons/mongo/Utils.scala
```diff
@@ -0,0 +1,14 @@
...
+package mongo
+
+import monix.eval.Task
+import monix.reactive.Observable
+import org.reactivestreams.Publisher
+
+object Utils {
+  implicit class PublisherOps[T](private val publisher: Publisher[T]) extends AnyVal {
+    def asMonix: Observable[T] = Observable.fromReactivePublisher(publisher)
+    def asMonixTask: Task[T] = Observable.fromReactivePublisher(publisher).firstL
```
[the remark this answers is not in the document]
ghik: Unused, will remove.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoDocumentFilter.scala
ghik: `Filters.exists("_id", false)` will not work for embedded documents.

Due to various random limitations of MongoDB queries I decided to write API only for things which are straightforward to express. Later we may think about emulating some lacking elements like "always false" filter.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoIndex.scala
ghik: These are passed separately in `IndexOptions`.

Yes, we could rewrite `MongoIndex` to better express various index types along with their options but I decided it would be too much work for now.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoDocumentFilter.scala
ddworak: That's fine by me, I'd just love to have a way to construct a generic method returning such `MongoDocumentFilter[T]` which is probably impossible on the library client side (without taking a ref / companion arg)

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -0,0 +1,292 @@
...
+import com.mongodb.client.result.{DeleteResult, UpdateResult}
+import com.mongodb.reactivestreams.client.{DistinctPublisher, FindPublisher, MongoCollection}
+import com.mongodb.{MongoNamespace, ReadConcern, ReadPreference, WriteConcern}
+import monix.eval.Task
+import monix.reactive.Observable
+import org.bson.codecs.configuration.CodecRegistry
+import org.bson.{BsonDocument, BsonValue}
+import org.reactivestreams.Publisher
+
+final class TypedMongoCollection[E <: BaseMongoEntity : MongoAdtFormat](
```
ddworak, line 16: You wouldn't expect me to say this, but I'd consider making this non-final. There are some cases where extending this with a bit of logic could serve as a simple yet complete persistence API, without further layers of composition.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -0,0 +1,292 @@
...
+    val options = setupOptions(new ReplaceOptions).upsert(upsert)
+    single(nativeCollection.replaceOne(filter.toBson, replacement, options))
+  }
+
+  def bulkWrite(
+    writes: Seq[MongoWrite[E]],
+    setupOptions: BulkWriteOptions => BulkWriteOptions = identity
+  ): Task[BulkWriteResult] = {
+    val requests = writes.iterator.map(_.toWriteModel).to[JList]
+    single(nativeCollection.bulkWrite(requests, setupOptions(new BulkWriteOptions)))
```
ddworak, line 274: You could consider skipping the call if `writes` is empty - the native driver notoriously throws exceptions in such case which is usually not what the user expects.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -0,0 +1,292 @@
...
+    upsert: Boolean = false, // extracted as separate param because it's very commonly used
+    setupOptions: ReplaceOptions => ReplaceOptions = identity
+  ): Task[UpdateResult] = {
+    val options = setupOptions(new ReplaceOptions).upsert(upsert)
+    single(nativeCollection.replaceOne(filter.toBson, replacement, options))
+  }
+
+  def bulkWrite(
+    writes: Seq[MongoWrite[E]],
+    setupOptions: BulkWriteOptions => BulkWriteOptions = identity
```
ddworak, line 271: How about an `ordered(false)` default? It's huge for performance, especially when sharding is involved.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -0,0 +1,292 @@
...
+  ): Task[BulkWriteResult] = {
+    val requests = writes.iterator.map(_.toWriteModel).to[JList]
+    single(nativeCollection.bulkWrite(requests, setupOptions(new BulkWriteOptions)))
+  }
+
+  def createIndex(
+    index: MongoIndex[E],
+    setupOptions: IndexOptions => IndexOptions = identity
```
ddworak, line 279: If we can afford to be opinionated, please add `background(true)`

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -0,0 +1,292 @@
...
+  def createIndexes(
+    indexes: Seq[(MongoIndex[E], IndexOptions)],
+    setupOptions: CreateIndexOptions => CreateIndexOptions = identity
+  ): Task[String] = {
+    val indexModels = indexes.iterator.map {
+      case (index, options) => new IndexModel(index.toBson, options)
+    }.to[JList]
+    single(nativeCollection.createIndexes(indexModels, setupOptions(new CreateIndexOptions)))
+  }
+}
```
ddworak, line 292: Would feel so good to have a nicer `.aggregate` API (or at least validate it fits with your model in the future). It's easily the ugliest, least safe Mongo code we have in our codebase.

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Embedded document types
+
+A type representing embedded document is very similar to an entity type, i.e. it must be:
+
+* a case class
+* a sealed trait/class with `@flatten` annotation
+
+However, an embedded document is not a toplevel entity and does not need an ID. Therefore, it doesn't need to
+extend `MongoEntity` and its companion must extend `MongoDataCompanion` rather than `MongoEntityCompanion`.
```
ddworak, line 135: Sometimes you really want to use an external type (even though it may be considered a bad practice). If it's possible to do that with this driver, maybe we could be less strict with the `must` and mention how to do this neatly.

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+For more examples, see the Scaladoc of the `.ref` macro or 
+[tests](https://github.com/AVSystem/scala-commons/blob/mongo-api/commons-mongo/src/test/scala/com/avsystem/commons/mongo/typed/MongoRefTest.scala).
+
+**NOTE**: some of the above references (e.g. unwrapping optional fields or narrowing sealed hierarchies) introduce
+an _implied filter_ that is automatically included into the query if this reference is used in a MongoDB query document
+or projection.
+
+`MongoPropertyRef` is a fundamental type that serves as an entry point for creating queries, projections, updates,
+sort orders, indices, etc.
+
```
ddworak, line 232: Would be nice to mention `MongoRef` and `MongoDataRef`, even if just to signal differences with this type.

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+For more examples, see the Scaladoc or
+[tests](https://github.com/AVSystem/scala-commons/blob/mongo-api/commons-mongo/src/test/scala/com/avsystem/commons/mongo/typed/MongoUpdateTest.scala).
+
+### `MongoIndex`
+
+A `MongoIndex[E]` represents a MongoDB index document for an entity of type `E`. Examples:
+
+* Single field index, e.g. `MyEntity.ref(_.int).ascendingIndex`
+* Multi-field index, e.g.
```
ddworak, line 287: They're usually called compound in the context of MongoDB

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+}
+
+import monix.execution.Scheduler.Implicits.global
+program.runToFuture
+```
+
+For more examples of database operations with `TypedMongoCollection`, see
+[tests](https://github.com/AVSystem/scala-commons/blob/mongo-api/commons-mongo/src/test/scala/com/avsystem/commons/mongo/typed/TypedMongoCollectionTest.scala).
+
+`TypedMongoCollection` exposes mostly the same operations as Reactive Streams `MongoCollection` but typed differently,
```
ddworak, line 334: This should be mentioned much earlier in these docs, after all the library doesn't make much sense without it :P

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+**NOTE**: You can think of `Observable` as an asynchronous version of `Iterable`.
+
+### Blocking
+
+It is recommended to use MongoDB API in a non-blocking way (by composing `Task`s and `Observables`s).
+However, sometimes this is not possible - some external API may force us into synchronous, blocking implementations.
+Then we have no choice but to wait for the database on the current JVM thread.
+
+In order to do this, use utilities provided by `com.avsystem.commons.concurrent.BlockingUtils` class. Your project should
+provide an implementation of this class (often a globally accessible object) where an appropriate Monix `Scheduler`s
```
ddworak, line 425: Can we **not** recommend using an `object` for this? Statically defined thread pools usually mean 0 runtime configuration for this.

docs/TypedMongo.md
```diff
@@ -0,0 +1,446 @@
...
+
+```scala
+import com.avsystem.commons.concurrent.BlockingUtils
+import monix.execution.Scheduler
+
+object Blocking extends BlockingUtils {
+  // some Scheduler usually reused throughout your application
+  def scheduler: Scheduler = ??? 
+  // some Scheduler usually reused throughout your application (unbounded, for blocking code)
+  def ioScheduler: Scheduler = ???
```
ddworak, line 436: Some general recommendations are important here.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/DataTypeDsl.scala
```diff
@@ -0,0 +1,207 @@
+package com.avsystem.commons
+package mongo.typed
+
+import com.avsystem.commons.annotation.{explicitGenerics, macroPrivate}
+import com.avsystem.commons.macros.serialization.MongoMacros
+
+trait DataRefDsl[E, T] {
```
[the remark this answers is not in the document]
ghik: I will rename `MongoDataRef` to `MongoToplevelRef`

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoFilter.scala
```diff
@@ -0,0 +1,152 @@
...
+  * @tparam T type of the filtered value
+  */
+sealed trait MongoFilter[T] {
+
+  import MongoFilter._
+
+  def on[E](prefix: MongoPropertyRef[E, T]): MongoDocumentFilter[E] =
+    PropertyValueFilter(prefix, this)
+
+  final def toBson: BsonDocument = this match {
```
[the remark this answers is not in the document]
ghik: For sealed hierarchies, I usually prefer pattern matching to abstract methods with implementations (more "functional" approach as opposed to "object oriented" one).

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoFormat.scala
```diff
@@ -0,0 +1,287 @@
...
+    elementFormat: MongoFormat[T]
+  ) extends MongoFormat[C[T]]
+
+  final case class Dictionary[M[X, Y] <: BMap[X, Y], K, V](
+    codec: GenCodec[M[K, V]],
+    keyCodec: GenKeyCodec[K],
+    valueFormat: MongoFormat[V]
+  ) extends MongoFormat[M[K, V]]
+
+  final case class Optional[O, T](
```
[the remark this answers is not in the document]
ghik: I agree but these types should not be used directly. If used directly, my intention was to always use them with prefix, i.e. `MongoFormat.Optional`, etc.

docs/TypedMongo.md
```diff
@@ -0,0 +1,447 @@
...
+
+A `MongoIndex[E]` represents a MongoDB index document for an entity of type `E`. Examples:
+
+* Single field index, e.g. `MyEntity.ref(_.int).ascendingIndex`
+* Multi-field index, e.g.
+
+  ```scala
+  import MongoIndexType._
+  val index: MongoIndex[MyEntity] =
+  MongoIndex(MyEntity.ref(_.str) -> Hashed, MyEntity.ref(_.int) -> Descending)
```
ddworak, line 292: This would actually be an error (hashed indexes cannot be part of a compound index). Maybe the API is too liberal?

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoFormat.scala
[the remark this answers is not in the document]
ddworak: If so, you could expose utility methods and encapsulate the types. It's not a huge issue, but there's quite a bit of "internal public" API in this PR

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoQueryOperator.scala
```diff
@@ -0,0 +1,78 @@
...
+  final case class Gt[T](value: T, format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class Gte[T](value: T, format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class Lt[T](value: T, format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class Lte[T](value: T, format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class In[T](values: Iterable[T], format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class Nin[T](values: Iterable[T], format: MongoFormat[T]) extends MongoQueryOperator[T]
+  final case class Exists[T](exists: Boolean) extends MongoQueryOperator[T]
+  final case class Type[T](bsonType: BsonType) extends MongoQueryOperator[T]
+  final case class Regex[T](pattern: String) extends MongoQueryOperator[T] //TODO: options
+  final case class Mod[T](divisor: Long, remainder: Long) extends MongoQueryOperator[T]
```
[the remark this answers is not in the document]
ghik: I tried using `com.mongodb.client.model.Filters` and it [failed](https://stackoverflow.com/questions/64557424/how-do-i-express-complex-elemmatch-mongodb-query-using-java-driver).

I believe that the compatibility layer you're talking about is the already available `Raw` operator.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoRef.scala
```diff
@@ -0,0 +1,329 @@
...
+/**
+  * Represents a reference to a particular "place" in a MongoDB document. The "place" may be an actual path
+  * inside the document ([[MongoPropertyRef]]) or the whole document _itself_ (you can think of it as an empty path).
+  *
+  * When the [[MongoRef]] points to the whole document, it may also narrow the type only to some subtype(s).
+  * See the [[DataRefDsl.as]] macro for more details on narrowing.
+  *
+  * @tparam E the data type
+  */
+sealed trait MongoRef[E, T] extends MongoProjection[E, T] with DataRefDsl[E, T] { self =>
```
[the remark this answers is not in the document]
ghik: I agree but also I don't like `MyRef.filter.is(value)` as opposed to `MyRef.is(value)`.
Is this any worse that the current situation with `BsonRef`?

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoRef.scala
```diff
@@ -0,0 +1,329 @@
...
+  /**
+    * Composes this reference with another one, effectively appending a "suffix" to this reference.
+    * This is conceptually similar to composing functions using `scala.Function1.andThen`.
+    */
+  def andThen[S](suffix: MongoRef[T, S]): suffix.ThisRef[E, S] = suffix compose this
+
+  def on[E0](ref: MongoRef[E0, E]): MongoProjection[E0, T] = compose(ref)
+}
+
+sealed trait MongoDataRef[E, T <: E] extends MongoRef[E, T] {
```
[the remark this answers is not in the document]
ghik: Sure, I'll add some but this type is not really used in the public API.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoRef.scala
```diff
@@ -0,0 +1,329 @@
...
+  * When the [[MongoRef]] points to the whole document, it may also narrow the type only to some subtype(s).
+  * See the [[DataRefDsl.as]] macro for more details on narrowing.
+  *
+  * @tparam E the data type
+  */
+sealed trait MongoRef[E, T] extends MongoProjection[E, T] with DataRefDsl[E, T] { self =>
+  def format: MongoFormat[T]
+  def projectionRefs: Set[MongoRef[E, _]] = Set(this)
+  def showRecordId: Boolean = false
+
```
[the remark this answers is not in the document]
ghik: This gets a bit complicated because `MongoRef` is technically a _partial_ function now. Also, for consistency I would probably have to add this on `MongoProjection` level and maybe `MongoFilter` should also have a memory-level filter, etc...

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: I understand but if you want me to maintain binary compat in a proper way, that might become annoying.

ddworak: That's up to you to decide then, I'm not expecting too much in terms of compatibility when I'm upgrading commons :D

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: Same remark as before. This would break the assumption that `TypedMongoCollection` is just a better-typed wrapper over the native collection, without changing semantics.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: Yeah, I haven't looked into it yet but I'm not sure if we can express pipelines in any better-typed way.

docs/TypedMongo.md
ghik: You want to use an external type as an _entity_ type or just _field_ type? As explained in this doc, you can use any type that has `GenCodec` as a field type (opaque).

ddworak: In my case it's field type (I also consider e.g. cross-compiled classes as external in this case). Just wanted to clarify that `must extend` is not really true.

docs/TypedMongo.md
ghik: These differences are explained in scaladoc. I don't want the tutorial to describe too many rarely used details like `MongoRef` and `MongoDataRef`. These don't really appear in everyday-use API.

commons-core/jvm/src/main/scala/com/avsystem/commons/concurrent/ObservableBlockingIterator.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import scala.concurrent.blocking
+import scala.concurrent.duration.TimeUnit
+
+/**
+  * An `Iterator` backed by a `BlockingQueue` backed by an `Observable`.
+  * This essentially turns an `Observable` into an `Iterator`
+  * (which requires blocking so use this only as a last resort).
+  */
+class ObservableBlockingIterator[T](
```
ddworak, line 20: I've seen an alternative implementation in RxJava `https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/operators/observable/BlockingObservableIterable.java`. You might want to take a look or borrow some tests for better coverage :eyes:

docs/TypedMongo.md
ghik: Well, I have actually used a globally accessible object for this...
I'm currently under impression that thread pools are one of the rare things that can be kept in globals because in order to  make sure that they are reused throughout your _entire process_. The whole point of using asynchronous API is to reduce the number of threads to minimum.

If we need compatibility then... JVM properties? (I would not think I'd ever say that).

docs/TypedMongo.md
ghik: `must extend` is true for an _embedded document_ whose internal structure is transparent. This doc section is about these.

docs/TypedMongo.md
ghik: Yeah I should rethink `MongoIndex`

commons-core/jvm/src/main/scala/com/avsystem/commons/concurrent/ObservableBlockingIterator.scala
ghik: Quite a different implementation though because this one isn't based on Monix

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: How do I create `BulkWriteResult` to represent that?

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoFormat.scala
```diff
@@ -0,0 +1,287 @@
...
+import com.avsystem.commons.misc.ValueOf
+import com.avsystem.commons.mongo.{BsonValueInput, BsonValueOutput}
+import com.avsystem.commons.serialization._
+import org.bson.{BsonNull, BsonValue}
+
+import scala.annotation.tailrec
+
+/**
+  * Typeclass that captures internal structure of a type that can be saved to MongoDB
+  * (directly as a toplevel entity or indirectly as an embedded value).
```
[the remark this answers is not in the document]
ghik: I'll add a section to the tutorial about this.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ddworak: You probably don't. Since you're wrapping the API, you can always introduce your own result type.

docs/TypedMongo.md
ddworak: I know it's tempting, but injecting some object which provides blocking capabilities **and** is a Scheduler / EC / TP is not that much of a hassle and makes (us) developers reconsider the way we run async tasks. Globally accessible object can also be messy in tests, it's very hard to control or mock it then.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/MongoFormat.scala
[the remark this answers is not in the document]
ghik: Yeah, but in order to get rid of this, I would have to add a lot of package-private modifiers. I would also prefer to leave a possibility to use this internal-ish API in corner cases.

docs/TypedMongo.md
ghik: I've removed the "globally accessible object" suggestion.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: But I want to expose `BulkWriteResult` since it contains useful information. Do you want me to wrap it into an `Opt` or something?

docs/TypedMongo.md
ghik: Well that's actually not true anymore: https://docs.mongodb.com/manual/core/index-hashed/#create-a-compound-hashed-index

ghik: I know that this API still allows creation of many invalid indexes but I think there's no point in trying to cover this with very precise static API - it will likely be complicated and become incompatible as MongoDB is evolving.

ghik:
@ddworak I think I've addressed all your concerns, either with code or with a reply. Some of these replies are in the spirit of "I'd prefer avoid changing this but I will if you think it's important".

docs/TypedMongo.md
ddworak: Whoa that's fresh. I'm okay with this being a bit liberal then, but I'm still not the biggest fan of this API in the sense that single indexes are created via extension methods on refs and compound indexes have a completely different feel (making them harder to discover I guess?). Maybe that's a good thing though given how different such indexes are.

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ddworak: I wouldn't mind that, basically yes.

docs/TypedMongo.md
```diff
@@ -444,3 +448,23 @@ Using this utility, you can:
...
+This old API provides:
+
+* `GenCodec` based serialization - four variants of `GenCodecCollection` creators for different native drivers
+* `BsonRef` - references to document properties, superseded by `MongoPropertyRef` in the new API
+* extension methods for raw `Bson` type for creating queries, updates, etc. - explicit imports are required
+
+In comparison to the old API, the new one provides:
+
+* `TypedMongoCollection` wrapper over Reactive Streams collection - as opposed to native `MongoCollection`s
+  * other drivers are unsupported because they are deprecated, synchronous or redundant
```
ddworak, line 466: > deprecated, synchronous or redundant

 I love it.

ddworak (approved):
LGTM, this is a significant improvement over previous approaches and fits nicely with our ecosystem. The leftover 3 comments are up to you to decide :shipit:

commons-mongo/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: I don't like this because it would force pattern matching or some other empty `Opt` handling in situations where the user is sure that `writes` are non-empty. Also, this goes against the "wrap with better API but don't change semantics" approach of `TypedMongoCollection`.

ghik:
Thanks a lot! I will merge this in current form in order to push things further in downstream projects.

PR #392, opened 2021-04-26
JsonStringInput can peek fields by name

ghik:
This makes it possible to read [flat sealed hierarchies](https://github.com/AVSystem/scala-commons/blob/master/docs/GenCodec.md#flat-format) even if `_case` field is not the first one.

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
@@ -191,8 +191,8 @@ class JsonStringInput(
 }
 
 final class JsonStringFieldInput(
-  val fieldName: String, reader: JsonReader, options: JsonOptions, objectInput: JsonObjectInput
-) extends JsonStringInput(reader, options, objectInput) with FieldInput
+  val fieldName: String, reader: JsonReader, options: JsonOptions, callback: AfterElement
+) extends JsonStringInput(reader, options, callback) with FieldInput
 
 final class JsonListInput(reader: JsonReader, options: JsonOptions, callback: AfterElement)
   extends ListInput with AfterElement {
@@ -224,6 +224,9 @@ final class JsonObjectInput(reader: JsonReader, options: JsonOptions, callback: 
 
   prepareForNext(first = true)
 
+  private[this] var peekIdx = if (end) -1 else reader.index
+  private[this] var peekedFields: MMap[String, Int] = _
+
   private def prepareForNext(first: Boolean): Unit = {
     reader.skipWs()
     end = reader.isNext('}')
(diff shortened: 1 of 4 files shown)


commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -224,6 +224,9 @@ final class JsonObjectInput(reader: JsonReader, options: JsonOptions, callback:
 
   prepareForNext(first = true)
 
+  private[this] var peekIdx = if (end) -1 else reader.index
+  private[this] var peekedFields: MMap[String, Int] = _
```
ddworak, line 228: `com.avsystem.commons.misc.CrossUtils#dictionary`? it's shared code after all

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
```diff
@@ -237,16 +240,65 @@ final class JsonObjectInput(reader: JsonReader, options: JsonOptions, callback:
...
+      else {
+        reader.pass(',')
+        reader.index
+      }
+    }
+
+    val alreadyPeeked = peekedFields.opt.flatMap(_.getOpt(name).map(idx => peekFieldInput(name, idx)))
+    alreadyPeeked orElse {
+      val savedIdx = reader.index
+      try {
```
ddworak, line 276: this try/catch/finally is not sufficiently tested, everything passes with this omitted

ghik: Well it's rather paranoic. I'm not sure if something can throw exceptions in there.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
```diff
@@ -384,4 +385,29 @@ class JsonStringInputOutputTest extends AnyFunSuite with SerializationTestUtils
...
+    assert(oi.peekField("b").map(GenCodec.read[Boolean]).contains(true))
+    assert(oi.nextField().fieldName == "a") // peeking should not affect `nextField`
+    assert(oi.peekField("d").map(GenCodec.read[Map[String, String]]).contains(Map.empty))
+    assert(oi.peekField("c").map(GenCodec.read[List[Int]]).contains(List(1, 2, 3)))
+    assert(oi.peekField("x").isEmpty)
+    assert(oi.peekField("a").map(GenCodec.read[Int]).contains(25))
+    assert(oi.nextField().fieldName == "b") // peeking should not affect `nextField`
+  }
+
+  test("reading flat sealed hierarchy with changed field order") {
```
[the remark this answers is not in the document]
ghik: Yeah, this is the result of incrementally adding stuff
* at first, there was no such thing as `peekField`, `Input`s & `Output`s were strictly required to keep the order
* `@outOfOrder` was added as a band-aid specifically for the fact that Mongo always returns `_id` as first field (otherwise keeping the field order properly)
* `peekField` was added then but it's not supported by all inputs/outputs so you can only rely on it when using specific input/output implementation

Overall, the initial design assumed that `Input` will _only_ read data that was written by its corresponding `Output`. So _technically_, `JsonStringInput` should not be used to read user data, e.g. in REST. Reality is, of course, different.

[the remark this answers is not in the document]
ddworak: Yeah, I guess it would make sense to add a general recommendation in gencodec.md about using `outOfOrder` vs peeking to avoid confusion.

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
ddworak: Seems odd that this part is the only one which gave you this feeling :D I'd probably opt for not doing it and finding out, but it's your call ;)

ddworak (approved):
LGTM with some leftover comments for you to resolve (or not)

commons-core/src/main/scala/com/avsystem/commons/serialization/json/JsonStringInput.scala
ghik: This part temporarily modifies state of `reader` and must restore after it's done.
Something like this automatically makes me write a `try`-`finally`.

commons-core/src/test/scala/com/avsystem/commons/serialization/json/JsonStringInputOutputTest.scala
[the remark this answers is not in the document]
ghik: Damn, I want to rewrite this entire doc. It's old.

PR #396, opened 2021-07-02
Observable extensions

ddworak:


commons-core/src/main/scala/com/avsystem/commons/concurrent/ObservableExtensions.scala
```diff
@@ -0,0 +1,46 @@
...
+        .toL(Array)
+        .map { arr =>
+          Sorting.stableSort(arr)
+          immutable.ArraySeq.unsafeWrapArray(arr)
+        }
+
+    def sortedByL[R](f: T => R)(implicit ord: Ordering[R], ct: ClassTag[T]): Task[ISeq[T]] =
+      sortedL(ord on f, ct)
+
+    def toL[R](factory: Factory[T, R]): Task[R] =
```
ghik, line 41: Can you add some scaladocs? This method's name may be especialy enigmatic for some people.

PR #397, opened 2021-07-05
Better CBOR support

ghik:
[Documentation](https://github.com/AVSystem/scala-commons/blob/better-cbor/docs/CBOR.md)


docs/CBOR.md
```diff
@@ -0,0 +1,139 @@
...
+keys and arbitrary CBOR-encoded keys. You can use this e.g. to give a numeric label to every textual field, reducing the
+size of resulting CBOR.
+
+```scala
+val keyCodec = new CborKeyCodec {
+  def writeFieldKey(fieldName: String, output: CborOutput): Unit = fieldName match {
+    case "intField" => output.writeInt(1)
+    case "strField" => output.writeInt(2)
+    case name => output.writeString(name)
+  }
```
WojciechKarpiel, line 83: This is an alternative to using `@cborKey` annotation, right?
Or is it more flexible?

docs/CBOR.md
```diff
@@ -0,0 +1,139 @@
...
+CBOR, which means two things:
+
+* if your case class contains a field typed as `Map[K, V]`, map keys will be serialized directly into CBOR rather than
+  converted into strings with `GenKeyCodec` (which would happen for standard `GenCodec`).
+* you can use `@cborKey` and `@cborDiscriminator` annotations in your case class & sealed hierarchy definitions
+
+Example:
+
+```scala
+@cborDiscriminator(0) sealed trait UnionData
```
WojciechKarpiel, line 117: Let me know if I understand this correctly:
The `@cborDiscriminator` and `@cborKey` aren't tied to each other, you don't have  to use `@cborKey` when using a `@cborDiscriminator`.

I was confused for a few seconds when I saw `@cborKey` used in every variant of the sealed class. I think the documentation would be clearer id these two concepts were presented separatedly, or one of the variant class didn't use the `@cborKey` annotation.

Also, what is the difference between using a `String` `cborKey` and the standard GenCodec annotation for changing name of a serialized field (can't remember it's name)?

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/CborAdtMetadata.scala
```diff
@@ -0,0 +1,245 @@
...
+  private class MappingCborKeyCodec(keyInfos: List[CborKeyInfo[_]]) extends CborKeyCodec {
+    val rawKeys: Map[String, RawCbor] =
+      keyInfos.mkMap(_.stringKey, _.rawKey)
+    val strKeys: Map[RawCbor, String] =
+      keyInfos.mkMap(_.rawKey, _.stringKey)
+
+    def writeFieldKey(fieldName: String, output: CborOutput): Unit =
+      rawKeys.getOpt(fieldName).fold(output.writeString(fieldName))(output.writeRawCbor)
+    def readFieldKey(input: CborInput): String =
+      input.readRawCbor() |> (cbor => strKeys.getOrElse(cbor, s"unknown_cbor_key:$cbor")) // unknown key will be ignored
```
WojciechKarpiel, line 224: unknown String keys are ignored as well, right?
Is there a flag to fail on unrecognized string fields? If yes, does it apply to CBOR as well?

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/CborOutput.scala
```diff
@@ -242,10 +236,13 @@ class CborListOutput(
 
 class CborObjectOutput(
   out: DataOutput,
-  fieldLabels: FieldLabels,
+  keyCodec: CborKeyCodec,
   sizePolicy: SizePolicy
 ) extends CborSequentialOutput(out, sizePolicy) with ObjectOutput {
 
+  private[this] var forcedKeyCodec: CborKeyCodec = _
+  private[this] def currentKeyCodec = if(forcedKeyCodec != null) forcedKeyCodec else keyCodec
```
WojciechKarpiel, line 244: What are the `forcedKeyCodec`s used for? Custom `@cborKey`s and `@cborDiscriminator`s? How do they differ from the `keyCoded` passed through the constructor?

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/RawCbor.scala
```diff
@@ -1,17 +1,42 @@
...
+
+import scala.annotation.tailrec
 
 final case class RawCbor(bytes: Array[Byte], offset: Int, length: Int) {
   require(offset >= 0 && length >= 0 && offset + length <= bytes.length)
 
+  override def equals(other: Any): Boolean = other match {
+    case RawCbor(otherBytes, otherOffset, otherLength) =>
+      @tailrec def loop(i: Int): Boolean =
+        i >= length || bytes(offset + i) == otherBytes(otherOffset + i) && loop(i + 1)
```
WojciechKarpiel, line 14: what about maps? Is it intentional that two equal maps before the serialization could in principle generate two non-equal byte arrays if the key ordering within a map is different?

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/RawCbor.scala
```diff
@@ -1,17 +1,42 @@
...
+      length == otherLength && loop(0)
+    case _ => false
+  }
+
+  override lazy val hashCode: Int = {
+    @tailrec def loop(i: Int, acc: Int): Int =
+      if (i >= offset + length) acc
+      else loop(i + 1, acc * 31 + bytes(i))
```
WojciechKarpiel, line 22: how will it behave on integer overflow? It is not specified by JVM IIRC

docs/CBOR.md
ghik: `CborKeyCodec` passed directly to `CborOutput`/`CborInput` is global - it will affect all the keys from all case classes, maps, etc. This is good e.g. for encoding SenML which has numeric labels for fields.

`@cborKey` applies only to some particular field of some particular case class (or discriminator value of a case class/object).

docs/CBOR.md
ghik: > The @cborDiscriminator and @cborKey aren't tied to each other, you don't have to use @cborKey when using a @cborDiscriminator.
Yes.

> I was confused for a few seconds when I saw @cborKey used in every variant of the sealed class. I think the documentation would be clearer id these two concepts were presented separatedly, or one of the variant class didn't use the @cborKey annotation.

Elaborate pls how you imagine it, I'm not sure what you mean (what do you mean by "variants", in particular?)

`@cborDiscriminator` is effectively a `@cborKey` for the generated discriminator field. Because the field is generated, there's no case class field to apply `@cborKey` to. Therefore a separate annotation is needed.

> Also, what is the difference between using a String cborKey and the standard GenCodec annotation for changing name of a serialized field (can't remember it's name)?

Effectively no difference except that `@cborKey` will only be honored by `CborInput`/`CborOutput` while `@name` works with every `Input`/`Output` implementation.

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/CborAdtMetadata.scala
ghik: There is no such feature in `GenCodec` itself. Unknown fields are ignored so that case classes may evolve in a backwards compatible way.

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/CborOutput.scala
ghik: The one passed through the constructor is inherited by nested outputs, the "forced" one is only for this particular `Output` instance.

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/RawCbor.scala
ghik: It will overflow. So what? This is literally the same as `java.util.Arrays#hashCode(byte[])` but I can't use it because it doesn't accept offsets.

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/RawCbor.scala
ghik: Is it a problem if you serialize the map to JSON string?

WojciechKarpiel: Well, yes, it is related to JSON aswell. Do you think this is not worth bothering?

docs/CBOR.md
WojciechKarpiel: by `variants` i mean the `case class`es that inherit from the `sealed trait`. What is the correct name?


> Elaborate pls how you imagine it
Something like this, so it doesn't look like there is a need to use "discriminator" and "field rename" `cborKeys` simultaneously
```
object UnionData extends HasCborCodec[UnionData] {
  @cborKey(1) case class Textual(@cborKey(1) txt: String) extends UnionData
  @cborKey(2) case class Numeric(@cborKey(true) num: Int) extends UnionData
  @cborKey(3) case class Mapping(map: Map[Int, String]) extends UnionData
}
```

Anyway, ofiginal is LGTM if you think this is inferior

WojciechKarpiel (approved):

commons-core/src/main/scala/com/avsystem/commons/serialization/cbor/RawCbor.scala
ghik: I think it's not because it's a fundamental "problem" for every serialization.

docs/CBOR.md
ghik: Ok, so the problem is that I shoved all the features in a single example at once?

WojciechKarpiel: Yes, at least it confused me (YMMV)

ghik: TBH it makes sense to set `@cborDiscriminator` if you're also setting `@cborKey`s so I wouldn't mind if people did things like in this example.

PR #398, opened 2021-07-16
Introduced `AutoOptionalParam` implicit as an alternative to `optionalParam` annotation

ghik:
Using implicit instances of `AutoOptionalParam` it is now possible to make some types behave as if there were always annotated with `@optionalParam` annotation. This is an opt-in mechanism, you have to bring implicit `AutoOptionalParam` instances explicitly into your scope, e.g. with `HasGenCodecWithDeps[AutoOptionalParams.type, ...]`

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/meta/OptionLike.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/meta/OptionLike.scala
@@ -52,3 +52,24 @@ object OptionLike {
   implicit def nOptOptionLike[A]: BaseOptionLike[NOpt[A], A] =
     new OptionLikeImpl(NOpt.Empty, NOpt.some, _.isDefined, _.get)
 }
+
+/**
+  * If there is an instance of [[AutoOptionalParam]] for some type `T` then all case class &
+  * RPC parameters of type `T` will be treated as if they were annotated with
+  * [[com.avsystem.commons.serialization.optionalParam @optionalParam]].
+  *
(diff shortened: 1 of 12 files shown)


ddworak:
Definitely useful and better than it was. Would be nice to be able to configure it globally at the input/output level, but that would require a change in their API and probably not solve your original use case for this? Guess I can still do that by adding some common companion which'd enable it for all my classes, which is good enough.

ghik:
@ddworak yes, that's the solution that doesn't require redesigning everything in and out. If there will ever be a next major version of `GenCodec` (e.g. for Scala 3) then this will most likely by the default behaviour.

ddworak (commented):
I'd add a test and a doc mention for the case with `AutoOptionalParam` present and the JSON / whatever still containing a "null" value, so it's obvious whether it's safe to enable this with legacy classes.

sebaciv (approved):

ghik:
@ddworak well, not really. The semantics of `@optionalParam` and `AutoOptionalParam` is that the field must be _missing_. I'll see if I can add some way to interpret `null`s as missing fields.

ghik:
@ddworak are you ok with the solution for accepting `null`s? This is technically an incompatible change of previous semantics where you could do something like `@optionalParam v: Opt[Opt[Thing]]` to capture both optionality and nullability.

ddworak (approved):
Thanks you for polishing this, LGTM!

PR #403, opened 2021-11-09
Reworked TypedMap and added support for it in typed Mongo API

ghik:


commons-core/src/main/scala/com/avsystem/commons/misc/TypedMap.scala
```diff
@@ -2,15 +2,30 @@ package com.avsystem.commons
 package misc
 
 import com.avsystem.commons.misc.TypedMap.GenCodecMapping
-import com.avsystem.commons.serialization.{GenCodec, GenKeyCodec, ObjectInput, ObjectOutput}
+import com.avsystem.commons.serialization._
 
 class TypedMap[K[_]](val raw: Map[K[_], Any]) extends AnyVal {
```
ddworak, line 7: Just a tiny bit of docs (maybe usage example?) would be helpful with adoption. All other examples are in tests, which don't get shipped to users.

ddworak (approved):
LGTM

PR #404, opened 2021-11-17
Add basic default transparent wrappers

ddworak:


commons-core/src/main/scala/com/avsystem/commons/serialization/TransparentWrapperCompanion.scala
```diff
@@ -43,3 +43,6 @@ abstract class TransparentWrapperCompanion[R, T] extends TransparentWrapping[R,
 abstract class StringWrapperCompanion[T] extends TransparentWrapperCompanion[String, T]
 abstract class IntWrapperCompanion[T] extends TransparentWrapperCompanion[Int, T]
 abstract class LongWrapperCompanion[T] extends TransparentWrapperCompanion[Long, T]
+abstract class FloatWrapperCompanion[T] extends TransparentWrapperCompanion[Float, T]
```
ghik, line 46: Are you srsly using `Float`?

ddworak: Not really, needed `Double` but added this for consistency.

PR #405, opened 2021-11-18
Convert optional value wrappers to immutable Iterable

bgrochal:
Requested change: `opt2iterable` implicit conversion returns `scala.collection.immutable.Iterable` instead of base `scala.collection.Iterable`.

Reason to change: implicit conversion is not applied when working with immutable APIs like `monix.reactive.Observable#flatMapIterable`. This change should not break backwards compatibility since `sci.Iterable` extends `sc.Iterable`. Also implementation is not affected since wrappers are converted to `List` underneath.

### DIFF
--- a/commons-core/src/main/scala/com/avsystem/commons/misc/NOpt.scala
+++ b/commons-core/src/main/scala/com/avsystem/commons/misc/NOpt.scala
@@ -1,5 +1,7 @@
 package com.avsystem.commons.misc
 
+import com.avsystem.commons.IIterable
+
 object NOpt {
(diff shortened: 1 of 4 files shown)


ghik:
It was done like this because that is how standard `Option` does it. I intended to keep `Opt`'s API as close to `Option` as possible - not sure exactly how to approach this.

ddworak:
I have a strange feeling we've hit that issue at least a couple of times before and never found any technical reason not to change this. My vote would be to find one or merge. Maybe we can even contribute the change to `Option` even if just to learn that we're wrong? :)

PR #409, opened 2022-02-01
Allow interpreting Int32 as Long when reading BSON values

sebaciv:


ghik (commented):
Can you also make it _write_ an INT32 when a `Long` value fits in an `Int`?

sebaciv:
I suppose it can be done, but I don't understand why? That would change DB representation of existing data after some update operation. IMO it's better if the type in DB corresponds to the type declared in the code. Change introduced by this PR is only for backward compatibility.

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
```diff
@@ -80,3 +80,8 @@ object SomethingComplex {
 
   implicit val codec: GenCodec[SomethingComplex] = GenCodec.materialize
 }
+
+case class SomethingLong(value: Long)
+object SomethingLong {
```
ghik, line 85: `HasGenCodec` ?

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonReaderInput.scala
```diff
@@ -16,7 +16,10 @@ class BsonReaderInput(br: BsonReader, override val legacyOptionEncoding: Boolean
   override def readString(): String = br.readString()
   override def readBoolean(): Boolean = br.readBoolean()
   override def readInt(): Int = br.readInt32()
-  override def readLong(): Long = br.readInt64()
+  override def readLong(): Long = {
```
ghik, line 19: Cosmetic: pls use pattern matching on `bsonType`

ghik (approved):

Issue #411, opened 2022-03-03
[TypedMongo] Allowing the discriminator field at any position


pulewicz:
What is the easiest way of reading class hierarchies with discriminator fields at the end of the object? Unfortunately, we need to adapt to an existing schema.

As explained in https://github.com/AVSystem/scala-commons/blob/master/docs/GenCodec.md#flat-sealed-hierarchy-format:

> It may also be sensitive to object field order because the discriminator field must be read before any other fields (this is only a problem with Input implementations that do not support random field access by name - JsonStringInput does not have this problem).

I did some checks, and it seems like `BsonReaderInput` has this problem, but `BsonValueInput` hasn't:  https://github.com/pulewicz/mongo-commons-playground/blob/master/src/main/scala/PlaygroundApp.scala#L47

Questions:

1. Is there any easy way to use `TypedMongoCollection` with value input and output? From what I see currently it's hardcoded and it will need creating your own versions of the following classes, is it correct? 
    
    - https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/GenCodecBasedBsonCodec.scala#L14
    - https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/core/GenCodecProvider.scala#L10
    - https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/core/GenCodecRegistry.scala#L14
    - https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala#L35

3. Is using the value input and output a good solution, or are there any disadvantages?

Regards,
Paweł Ulewicz.

ghik:
Hi,

Thanks for the report. I'll see if I can make `BsonReaderInput` independent of field order similarly to how it's done in `JsonStringInput`. Otherwise I will look for ways make it possible to use `BsonValueInput` in `TypedMongoCollection`. The primary advantage of `BsonReaderInput` over `BsonValueInput` is performance - it avoids the intermediate `BsonValue` representation.

pulewicz:
Hi,

Great to hear that and thanks for help. Please contact me if you need more information.

Regards,
Paweł Ulewicz.

ghik:
Fixed in 2.6.0

[...]

Issue #410, opened 2022-03-03
Mapping scala BigDecimal to BSON NumberDecimal instead of binary


pulewicz:
What is the best to map `scala.math.BigDecimal` to `NumberDecimal` instead of binary? 

I created a repo with this example and currently it looks like that:

```
case class TypeWithBigDecimal(x: BigDecimal, y: BigDecimal)

TypeWithBigDecimal(x = 103234.42343, y = 12344.423)

{"x": {"$binary": {"base64": "AmdTOqcAAAAF", "subType": "00"}}, "y": {"$binary": {"base64": "ALxcZwAAAAM=", "subType": "00"}}}
```

https://github.com/pulewicz/mongo-commons-playground/blob/master/src/main/scala/PlaygroundApp.scala#L89

From what I see, this is currently hardcoded in all of of the Bson outputs, so I guess the only way would be to create your own output implementation, isn't it?

https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonValueOutput.scala#L39
https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonWriterOutput.scala#L26
https://github.com/AVSystem/scala-commons/blob/master/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonWriterOutput.scala#L63

Are you planning to make it configurable?

Regards,
Paweł Ulewicz.

ghik:
Hi,

The problem with `NumberDecimal` is that it's not actually _decimal_ in the same way as Scala's `BigDecimal` is. In fact, the `NumberDecimal` (or `org.bson.types.Decimal128` in Java representation) is a 128-bit _floating point_ representation. This means that its exponent expresses a power of 2 rather than a power of 10. Additionally, `BigDecimal` can be of arbitrary size while `Decimal128` is obviously limited to 128 bits.

The consequence is that some values of `BigDecimal` cannot be precisely represented as `Decimal128`, e.g. the number `0.1` can be represented just fine as `BigDecimal` but representing it with `Decimal128` will lose some precision.

As a workaround, you can use `Decimal128` values directly instead of `BigDecimal`, e.g.

```scala
import com.avsystem.commons.serialization._
import com.avsystem.commons.mongo._
import org.bson.types.Decimal128

case class TypeWithDecimals(x: Decimal128, y: Decimal128)
[...]

PR #412, opened 2022-03-03
Changed `BsonInput`/`BsonOutput` implementations to use optimized value representations

ghik:
* When writing numbers, `BsonOutput` implementations try to use smallest possible type
  * e.g. when writing a `Long` that actually fits into an `Int`, value will be written as `BsonInt32`
* `Decimal128` is used for `BigInt` and `BigDecimal` where possible
* `BsonInput` representations are lenient when reading numbers and accept multiple `BsonType`s

This change should not break backwards compatibility because:
* `BsonInput` implementations have been updated to be lenient
* Having mixed types for the same field (e.g. `BsonInt32` and `BsonInt64`) is not a problem for MongoDB queries and indexes (value comparisons)

This also addresses #410 

Additional piggybacked changes: `peekField` implementation for `BsonReaderInput`, addressing #411

### DIFF
--- a/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
+++ b/commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
@@ -1,11 +1,12 @@
 package com.avsystem.commons
 package mongo
 
-import java.nio.ByteBuffer
-
+import com.avsystem.commons.serialization.GenCodec.ReadFailure
 import com.avsystem.commons.serialization.{InputAndSimpleInput, InputMetadata, OutputAndSimpleOutput, TypeMarker}
-import org.bson.BsonType
+import org.bson.{BsonInvalidOperationException, BsonType}
 import org.bson.types.{Decimal128, ObjectId}
+
+import java.nio.ByteBuffer
 
 object ObjectIdMarker extends TypeMarker[ObjectId]
 object Decimal128Marker extends TypeMarker[Decimal128]
@@ -17,6 +18,48 @@ trait BsonInput extends Any with InputAndSimpleInput {
   def readDecimal128(): Decimal128
 
   protected def bsonType: BsonType
+
+  protected def handleFailures[T](expr: => T): T =
+    try expr catch {
+      case e: BsonInvalidOperationException => throw new ReadFailure(e.getMessage, e)
+    }
+
+  protected def wrongType(expected: BsonType*): Nothing =
+    throw new ReadFailure(s"Encountered BsonValue of type $bsonType, expected ${expected.mkString(" or ")}")
+
+  protected def expect[T](tpe: BsonType, value: => T): T =
+    if (bsonType == tpe) handleFailures(value)
+    else wrongType(tpe)
+
+  override def readBigInt(): BigInt = handleFailures {
+    bsonType match {
+      case BsonType.INT32 | BsonType.INT64 =>
+        BigInt(readLong())
+      case BsonType.DECIMAL128 =>
+        val bigDec = BigDecimal(readDecimal128().bigDecimalValue())
+        bigDec.toBigIntExact.getOrElse(throw new ReadFailure(s"whole number expected but got $bigDec"))
+      case BsonType.BINARY =>
+        BigInt(readBinary())
+      case _ =>
+        wrongType(BsonType.INT32, BsonType.INT64, BsonType.DECIMAL128, BsonType.BINARY)
+    }
+  }
+
+  override def readBigDecimal(): BigDecimal =
+    handleFailures {
+      bsonType match {
+        case BsonType.INT32 | BsonType.INT64 =>
+          BigDecimal(readLong())
+        case BsonType.DOUBLE =>
+          BigDecimal(readDouble())
+        case BsonType.DECIMAL128 =>
+          BigDecimal(readDecimal128().bigDecimalValue)
+        case BsonType.BINARY =>
+          BsonInput.bigDecimalFromBytes(readBinary())
+        case _ =>
+          wrongType(BsonType.INT32, BsonType.INT64, BsonType.DOUBLE, BsonType.DECIMAL128, BsonType.BINARY)
+      }
+    }
 
   override def readMetadata[T](metadata: InputMetadata[T]): Opt[T] =
     metadata match {
@@ -47,6 +90,19 @@ trait BsonOutput extends Any with OutputAndSimpleOutput {
   def writeObjectId(objectId: ObjectId): Unit
   def writeDecimal128(decimal128: Decimal128): Unit
 
+  override def writeBigInt(bigInt: BigInt): Unit =
+    if (bigInt.isValidLong) writeLong(bigInt.longValue)
+    else Decimal128Utils.fromBigDecimal(BigDecimal(bigInt)) match {
+      case Opt(dec128) => writeDecimal128(dec128)
+      case Opt.Empty => writeBinary(bigInt.toByteArray)
+    }
+
+  override def writeBigDecimal(bigDecimal: BigDecimal): Unit =
+    Decimal128Utils.fromBigDecimal(bigDecimal) match {
+      case Opt(dec128) => writeDecimal128(dec128)
+      case Opt.Empty => writeBinary(BsonOutput.bigDecimalBytes(bigDecimal))
+    }
+
   override def keepsMetadata(metadata: InputMetadata[_]): Boolean =
     BsonTypeMetadata == metadata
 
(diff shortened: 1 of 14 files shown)


commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
```diff
@@ -54,13 +116,15 @@ trait BsonOutput extends Any with OutputAndSimpleOutput {
...
       case _ => false
     }
 }
 
 object BsonOutput {
   def bigDecimalBytes(bigDecimal: BigDecimal): Array[Byte] = {
     val unscaledBytes = bigDecimal.bigDecimal.unscaledValue.toByteArray
-    ByteBuffer.allocate(unscaledBytes.length + Integer.BYTES).put(unscaledBytes).putInt(bigDecimal.scale).array
+    ByteBuffer.allocate(unscaledBytes.length + Integer.BYTES)
+      .put(unscaledBytes).putInt(bigDecimal.scale).array
```
ddworak, line 128: Not a part of this change, but sad that we have to create a byte[] array just to copy it to a slightly larger one. I assume this intermediate step could be avoided

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonReaderInput.scala
```diff
@@ -1,48 +1,76 @@
...
+    bsonType == BsonType.NULL && {
       br.readNull()
       true
     }
 
-  override def readString(): String = br.readString()
-  override def readBoolean(): Boolean = br.readBoolean()
-  override def readInt(): Int = br.readInt32()
+  override def readString(): String =
+    expect(BsonType.STRING, br.readString())
```
ddworak, line 21: Wouldn't these type expectations be validated by `org.bson.AbstractBsonReader#verifyBSONType` when `BsonReader#readXXX` is called anyway? The current tests don't mind me removing `expect` completely, but I assume you'd rather `handleFailures` everywhere.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonReaderInput.scala
```diff
@@ -63,7 +91,48 @@ final class BsonReaderListInput(it: BsonReaderIterator[BsonReaderInput]) extends
...
+  )
+
+  private[this] var peekMark: BsonReaderMark = br.getMark
+  private[this] var peekedFields: MHashMap[String, BsonValue] = _
+
+  override def peekField(name: String): Opt[BsonFieldInput] =
+    br match {
+      case _: BsonDocumentReader =>
+        // Looks like there's a bug in BsonDocumentReader.Mark implementation, tests don't pass
+        // TODO: find and report/fix this bug
```
ddworak, line 110: Don't forget to do it :D

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
```diff
@@ -33,7 +33,10 @@ object SomethingPlain extends HasGenCodec[SomethingPlain] {
     string <- arbitrary[String]
     boolean <- arbitrary[Boolean]
     int <- arbitrary[Int]
-    long <- arbitrary[Long]
+    long <- Gen.oneOf(
+      Gen.choose(Long.MinValue, Int.MinValue - 1L),
+      Gen.choose(Int.MaxValue + 1L, Long.MaxValue),
+    )
```
ddworak, line 39: Why skip the "boring" integers?

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
```diff
@@ -33,7 +33,10 @@ object SomethingPlain extends HasGenCodec[SomethingPlain] {
...
     timestamp <- arbitrary[JDate]
     double <- arbitrary[Double]
     binary <- Gen.buildableOf[Array[Byte], Byte](arbitrary[Byte]).map(new Bytes(_))
```
ddworak, line 42: You could add some of the decimal / big types changed within the scope of this PR.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
```diff
@@ -1,23 +1,68 @@
...
+        BigInt(readLong())
+      case BsonType.DECIMAL128 =>
+        val bigDec = BigDecimal(readDecimal128().bigDecimalValue())
+        bigDec.toBigIntExact.getOrElse(throw new ReadFailure(s"whole number expected but got $bigDec"))
+      case BsonType.BINARY =>
+        BigInt(readBinary())
+      case _ =>
+        wrongType(BsonType.INT32, BsonType.INT64, BsonType.DECIMAL128, BsonType.BINARY)
+    }
+  }
```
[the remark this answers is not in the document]
ghik: Elaborate pls, I'm calling `readLong()` here to reuse the code for `Long` as much as possible. I don't see what I can do more.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
ghik: How?

I can only imagine it if `Input` & `Output` used some kind of `ByteString` instead of raw array but that would be a heavier and breaking change (also, likely to introduce external deps).

ddworak: Copy and modify `toByteArray` logic, use own array and `ByteBuffer.wrap` it if you need to use it as such.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonReaderInput.scala
ghik: Kind of yes but sometimes I want to call `expect` with multiple BSON types and this will not be covered by `verifyBSONType` (with the error message that I want). Therefore I use `expect` everywhere for consistency.

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
ghik: Because this test assumes that all of them will be `BsonInt64` which is no longer true (in general) after my changes.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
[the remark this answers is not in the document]
ddworak: That won't be obvious in the future. At worst add a comment, at best update the test please.

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
ghik: Yeah but this particular test assumes very specific BSON serialization which is now more complicated (same problem as with your previous comment on arbitrary `Long`s).

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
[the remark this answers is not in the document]
ddworak: As mentioned, I only care about future consistency here.

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
ddworak: That won't be obvious in the future. At worst add a comment, at best update the test please.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonReaderInput.scala
```diff
@@ -1,48 +1,76 @@
...
+  override def readString(): String =
+    expect(BsonType.STRING, br.readString())
+
+  override def readBoolean(): Boolean =
+    expect(BsonType.BOOLEAN, br.readBoolean())
+
+  override def readInt(): Int =
+    expect(BsonType.INT32, br.readInt32())
+
   override def readLong(): Long = bsonType match {
```
ddworak, line 29: Missing `handleFailures`?

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
ghik: Is it ok to not give a damn about this in this review?

commons-mongo/jvm/src/test/scala/com/avsystem/commons/mongo/SomethingPlain.scala
ghik: Update how? I would just remove `Long`s from this test altogether.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/BsonInputOutput.scala
ddworak: Your code didn't make anything worse for sure, so that's not a blocker on my side.

sebaciv (approved):

PR #413, opened 2022-03-18
Accept ints as doubles in Bson inputs

pulewicz:
Just a draft, I'm not sure it's a good approach.


ghik:
Hi,

I'm fine with reading `INT32` as `Double` since every `Int` value fits into a `Double`. This is not the case with `INT64` though - reading it as `Double` is not safe.

There are two approaches we can take here:
* only allow `DOUBLE` and `INT32` BSON types in `readDouble()`
* allow other types like `INT64` and `DECIMAL128` but validate whether the value is a valid `Double`, throw exception otherwise

If sufficient, the first approach would be preferable by me because it's much simpler.

Thanks,
Roman

pulewicz:
I'm also fine with the first approach. I updated the code to disallow `Int64`, I also added a test for `BsonBinaryReader`.

pulewicz:
I'm also fine with the first approach. I updated the code to disallow `Int64`, I also added a test for `BsonBinaryReader`.

PR #417, opened 2022-05-25
Typed Mongo API transaction support

ghik:
Apart from transactions themselves, the entire API has been expanded with more `Typed*` wrappers:
* `TypedMongoClient`
* `TypedMongoDatabase`
* `TypedClientSession`

This way it is possible use the typed MongoDB API without directly using any classes from the ReactiveStreams driver.


commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -90,24 +94,29 @@ class TypedMongoCollection[E <: BaseMongoEntity](
...
   def drop(): Task[Unit] =
-    empty(nativeCollection.drop())
+    empty(optionalizeFirstArg(nativeCollection.drop(sessionOrNull)))
 
   def renameCollection(
     namespace: MongoNamespace,
     setupOptions: RenameCollectionOptions => RenameCollectionOptions = identity
   ): Task[Unit] =
-    empty(nativeCollection.renameCollection(namespace, setupOptions(new RenameCollectionOptions)))
+    empty(optionalizeFirstArg(
```
wodomierz, line 103: maybe use following format ? :
```
    empty(optionalizeFirstArg(
      nativeCollection.renameCollection(sessionOrNull, namespace, setupOptions(new RenameCollectionOptions))
    ))
```

wodomierz: if yes, then apply the rule also to other method calls of such kind.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
```diff
@@ -0,0 +1,108 @@
...
+  ): TypedMongoClient =
+    new TypedMongoClient(MongoClients.create(settings, driverInformation))
+}
+
+/**
+  * A better-typed wrapper over [[MongoClient]]. Uses Monix [[Task]] and [[Observable]] instead of
+  * [[org.reactivestreams.Publisher]]. Returns similar better-typed wrappers for database and client session objects.
+  */
+class TypedMongoClient(
+  val nativeClient: MongoClient,
```
wodomierz, line 46: Maybe make nativeClient and clientSession private? I don't see any reason to make them public. On the other hand, in the future, making public from private will be possible, while the opposite won't.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -14,11 +14,25 @@ import org.bson.codecs.configuration.CodecRegistry
...
 import org.reactivestreams.Publisher
 
-class TypedMongoCollection[E <: BaseMongoEntity](
-  rawCollection: MongoCollection[_]
-)(
-  implicit meta: MongoEntityMeta[E]
-) extends DataTypeDsl[E] {
+class TypedMongoCollection[E <: BaseMongoEntity] private(
+  /** The native (Reactive Streams driver) collection */
+  val nativeCollection: MongoCollection[E],
```
wodomierz, line 19: as above: private fields ?

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedClientSession.scala
```diff
@@ -0,0 +1,103 @@
...
+import com.mongodb.{ClientSessionOptions, ServerAddress, TransactionOptions}
+import monix.eval.Task
+import org.bson.{BsonDocument, BsonTimestamp}
+
+import java.io.Closeable
+
+/**
+  * Better typed wrapper over [[ClientSession]].
+  */
+class TypedClientSession(val nativeSession: ClientSession)
```
wodomierz, line 16: as below, a I am not sure if public val is a good option here.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
ghik: I want them public - the `Typed*` API is incomplete and I want to give the user a fallback option. This is especially true with `TypedMongoCollection` itself.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
ghik: as above: I want public

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -90,24 +94,29 @@ class TypedMongoCollection[E <: BaseMongoEntity](
...
-    single(nativeCollection.countDocuments(filter.toBson, setupOptions(new CountOptions))).asInstanceOf[Task[Long]]
+    single(optionalizeFirstArg(
+      nativeCollection.countDocuments(sessionOrNull, filter.toBson, setupOptions(new CountOptions))
+    )).asInstanceOf[Task[Long]]
 
   def estimatedDocumentCount(
     setupOptions: EstimatedDocumentCountOptions => EstimatedDocumentCountOptions = identity
   ): Task[Long] =
-    single(nativeCollection.estimatedDocumentCount(setupOptions(new EstimatedDocumentCountOptions))).asInstanceOf[Task[Long]]
+    single(optionalizeFirstArg(
```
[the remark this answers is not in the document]
ghik: No I just forgot to pass `sessionOrNull` - good catch.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -14,11 +14,25 @@ import org.bson.codecs.configuration.CodecRegistry
...
+class TypedMongoCollection[E <: BaseMongoEntity] private(
+  /** The native (Reactive Streams driver) collection */
+  val nativeCollection: MongoCollection[E],
+  docCollection: MongoCollection[BsonDocument],
+  val clientSession: Opt[TypedClientSession],
+)(implicit
+  meta: MongoEntityMeta[E]
+) extends DataTypeDsl[E] with TypedMongoUtils {
+
+  def this(
```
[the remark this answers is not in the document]
ghik: With current API, the idiomatic way to obtain `TypedMongoCollection` is by calling `TypedMongoDatabase.getCollection` which in turn is obtained by calling `TypedMongoClient.getDatabase` which has a companion object with `apply`.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
```diff
@@ -0,0 +1,108 @@
...
+    driverInformation: MongoDriverInformation,
+  ): TypedMongoClient =
+    new TypedMongoClient(MongoClients.create(settings, driverInformation))
+}
+
+/**
+  * A better-typed wrapper over [[MongoClient]]. Uses Monix [[Task]] and [[Observable]] instead of
+  * [[org.reactivestreams.Publisher]]. Returns similar better-typed wrappers for database and client session objects.
+  */
+class TypedMongoClient(
```
wodomierz, line 45: Observations:
1. in the beginning I was a little bit confused with the layers in this code. 
This 2 methods: withSession and inTransaction are IMO from different "layouts" (calling inTransaction, which creates new session in class, which can have other session is a little bit weird).
However it is consistent with the native mongo api, so it is ok for me. 
2. I am wondering if there is some option, to automate adding withSession, when calling inTransaction. Maybe something like this:
```
  def inTransaction[T](
    sessionOptions: ClientSessionOptions = ClientSessionOptions.builder().build(),
    transactionOptions: TransactionOptions = TransactionOptions.builder().build(),
  )(
    task: this.type => Task[T]
  ): Task[T] = ???
  ```
  however I am not convinced about this - I am just sharing thoughts.

Comment (ghik) [reply commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala:45]: I already wrote `TypedMongoClient.inTransaction`, is that what you mean?

Comment (wodomierz) [reply commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala:45]: No, TypedMongoClient.inTransaction requires to additionally write withSession. I am thinking how to serve less-general code, but with less requirements.

File: commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala (hunk 6 above)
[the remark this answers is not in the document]
Comment (wodomierz) [reply commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala:26]: I know, nevertheless it is public method, so it can be created also by calling def this(). I am just not a big fan of `def this()` .

**Review by wodomierz — APPROVED** [review]

File: commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala [hunk 8]
```diff
@@ -0,0 +1,108 @@
...
+  def listDatabaseNames: Observable[String] =
+    multi(optionalizeFirstArg(nativeClient.listDatabaseNames(sessionOrNull)))
+
+  def listDatabases: Observable[Document] =
+    multi(optionalizeFirstArg(nativeClient.listDatabases(sessionOrNull)))
+
+  def listDatabases[T: GenCodec]: Observable[T] =
+    listDatabases.map(doc => BsonValueInput.read[T](doc.toBsonDocument))
+
+  //TODO: `watch` methods
```
ddworak, line 66: That'd actually be very cool (outside of the scope of this PR), we have a couple of usages.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoCollection.scala
```diff
@@ -14,11 +14,25 @@ import org.bson.codecs.configuration.CodecRegistry
...
-  implicit meta: MongoEntityMeta[E]
-) extends DataTypeDsl[E] {
+class TypedMongoCollection[E <: BaseMongoEntity] private(
+  /** The native (Reactive Streams driver) collection */
+  val nativeCollection: MongoCollection[E],
+  docCollection: MongoCollection[BsonDocument],
+  val clientSession: Opt[TypedClientSession],
+)(implicit
+  meta: MongoEntityMeta[E]
+) extends DataTypeDsl[E] with TypedMongoUtils {
```
ddworak, line 24: nit: breaking redundant or trailing comma missing :D

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def startSession(
+    options: ClientSessionOptions = ClientSessionOptions.builder().build()
+  ): Task[TypedClientSession] =
+    single(nativeClient.startSession(options)).map(new TypedClientSession(_))
```
ddworak, line 71: This pattern with "storing" the session in the client will discourage easy sharing of `TypedMongoClient` in DAO classes etc. I'm not saying it's wrong, it may just be tricky for the users - with transactions you will most often want to use a single native client and a `TypedMongoClient` per transaction.

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+    * Executes some code in context of a MongoDB client session. The session is closed afterwards.
+    *
+    * Note: in order for actual MongoDB operations to be associated with the session, you need to use
+    * `withSession` on [[TypedMongoClient]], [[TypedMongoDatabase]] or [[TypedMongoCollection]] and use the
+    * returned copy of these objects.
+    */
+  def inSession[T](
```
ddworak, line 80: https://www.mongodb.com/docs/manual/core/read-isolation-consistency-recency/#client-sessions-and-causal-consistency-guarantees

`Applications must ensure that only one thread at a time executes these operations in a client session.`

Can we ensure that at the driver level? ClientSession is `NotThreadSafe` and there are in fact issues with it when any parallelism is involved (we tested).

ghik: As far as I understand this is only a problem when someone intentionally uses a client session in parallel. Using it in subsequent but asynchronous invocations should be fine.

I don't immediately see a simple way to secure this. What do you have in mind?

commons-mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/TypedMongoClient.scala
ddworak: I'm okay with this, it's not worse than existing driver API.

ddworak (approved):

PR #419, opened 2022-06-28
Blocking Either as IO in BlockingUtils

wodomierz:


commons-core/jvm/src/main/scala/com/avsystem/commons/concurrent/BlockingUtils.scala
```diff
@@ -38,6 +39,10 @@ abstract class BlockingUtils {
   def asTask[T](blockingCode: => T): Task[T] =
     Task.eval(blockingCode).executeOn(ioScheduler, forceAsync = true)
 
+  def asIO[A, B](blockingCode: => Either[A, B]): IO[A, B] = IO.deferTotal {
```
ghik, line 42: Can you please debrace this into one line?

commons-core/jvm/src/main/scala/com/avsystem/commons/concurrent/BlockingUtils.scala
```diff
@@ -38,6 +39,10 @@ abstract class BlockingUtils {
...
+    IO.fromEither(blockingCode)
+  }.executeOn(ioScheduler, forceAsync = true)
```
ghik, line 44: `forceAsync = true` is the default AFAIR

wodomierz: I know, I just want to be consistent with the asTask impl.

commons-core/jvm/src/main/scala/com/avsystem/commons/concurrent/BlockingUtils.scala
wodomierz: done

ghik (approved):

PR #428, opened 2023-01-13
proper support for transparent wrappers in typed mongo

ghik:


ddworak:
Should we care about differences in handling transparent wrappings and `@transparent` annotation?

ghik:
@ddworak yes, that irks me too and I was thinking whether I should unify how `MongoAdtFormat` represents case classes with `@transparent` and types with `TransparentWrapping`.

In the end I decided not to do it because:
* `@transparent` is somewhat deprecated, you should favor `TransparentWrapping` where possible
* `@transparent` requires that the type is a case class (-ish) so in the end it is accurate to represent its format with `MongoAdtFormat.RecordFormat`

ddworak (approved):
Understandable. Let's ship it :ship:

PR #429, opened 2023-01-30
Fix implicit resolution conflict when MongoPolyDataCompanion is used with custom implicits

ghik:
`MongoPolyAdtInstances.codecFromFormat` was a bit too broad and conflicted with implicits injected via `MacroInstances`.

This is a rather obscure scenario, please see the added test which shows the minimized problem.

### DIFF
--- a/commons-mongo/jvm/src/main/scala-2.13/com/avsystem/commons/mongo/typed/MongoPolyDataCompanion.scala
+++ b/commons-mongo/jvm/src/main/scala-2.13/com/avsystem/commons/mongo/typed/MongoPolyDataCompanion.scala
@@ -12,13 +12,12 @@ trait MongoPolyAdtInstances[D[_]] {
   def codec[T: GenCodec]: GenObjectCodec[D[T]]
 
   /**
-    * We need to accept an implicit `GenObjectCodec[D[T]]` because materialization of
-    * [[MongoAdtFormat]] requires it ([[MongoAdtFormat.codec]]). In practice, it can be derived
-    * from the `MongoFormat[T]` that is already accepted by this method but we don't want to do this in this
-    * trait to avoid problems with priorities of implicits. Because of that, this implicit is actually constructed
-    * by [[AbstractMongoPolyDataCompanion.format]].
+    * We need to accept an implicit `GenCodec[T]` because materialization of
+    * [[MongoAdtFormat]] requires a [[GenObjectCodec]] ([[MongoAdtFormat.codec]]). In practice, it can be derived
+    * from the `MongoFormat[T]` that is already accepted by this method but we have to be careful about priority of
+    * implicits. Because of that, this implicit is actually provided by [[AbstractMongoPolyDataCompanion.format]].
     */
-  def format[T: MongoFormat](implicit codec: GenObjectCodec[D[T]]): MongoAdtFormat[D[T]]
+  def format[T: MongoFormat : GenCodec]: MongoAdtFormat[D[T]]
 }
 
 abstract class AbstractMongoPolyDataCompanion[Implicits, D[_]](implicits: Implicits)(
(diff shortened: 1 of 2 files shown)


commons-mongo/jvm/src/main/scala-2.13/com/avsystem/commons/mongo/typed/MongoPolyDataCompanion.scala
```diff
@@ -1,26 +1,35 @@
...
+import com.avsystem.commons.annotation.{bincompat, explicitGenerics}
 import com.avsystem.commons.meta.MacroInstances
 import com.avsystem.commons.mongo.BsonGenCodecs
 import com.avsystem.commons.serialization.{GenCodec, GenObjectCodec}
 
 import scala.annotation.compileTimeOnly
 
 trait MongoPolyAdtInstances[D[_]] {
-  // needed by MongoAdtFormat.materialize for generic type
-  protected final implicit def codecFromFormat[T: MongoFormat]: GenCodec[T] = MongoFormat[T].codec
```
mikegpl, line 13: I'm not familiarized well with this type structure, so here might go stupid question - the issue was that this was `protected` and it conflicted with `GenCodec` from the supplied `MacroInstances`?

ghik: It goes a bit deeper than that.

What happens here is:
* The `MongoPolyAdtInstances.format` method is implemented by the macro `MongoAdtFormat.materialize` (it happens via the machinery of `MacroInstances` - read its scaladoc if you want to know more)
* Materialization of `MongoAdtFormat[T]` requires an implicit `GenObjectCodec[T]` instance (as defined by `MongoAdtFormat` itself)
* This `GenObjectCodec[T]` instance can be easily provided as `MongoPolyAdtInstances.codec[T](MongoFormat[T].codec)` but there is no implicit that does this
* Therefore, I added the `codecFromFormat` implicit that does it (in conjunction with `codec` method from the data type's companion)
* Unfortunately, `codecFromFormat` started conflicting with other implicits injected via `MacroInstances` machinery
* Therefore I just added the required implicit `GenCodec[T]` to the `format` method - which has more limited scope than the previously over-generic `codecFromFormat` method.

Yes, this isn't pretty - type-parameterized data types can be tricky with typeclass instance materialization. Sometimes it may be best to avoid them completely. Fortunately, they are easily avoidable in DB-related code if things turn out to be too complicated (at the price of some additional boilerplate).

ghik:
Note: I decided to ignore binary compatibility as there is literally 0 chance this quite obscure feature is used in any _library_ code.

commons-mongo/jvm/src/main/scala-2.13/com/avsystem/commons/mongo/typed/MongoPolyDataCompanion.scala
mikegpl: Thanks for the comprehensive explanation!

mikegpl (approved):
LGTM, could you take a look @ddworak? This blocks us from upgrading IAM to the latest version in CDM

commons-mongo/jvm/src/test/scala-2.13/com/avsystem/commons/mongo/typed/PolyDataWithCustomImplicits.scala
```diff
@@ -0,0 +1,23 @@
...
+  implicit val customWrappyCodec: GenCodec[CustomWrappy] =
+    GenCodec.transformed[CustomWrappy, String](_.value, CustomWrappy)
+}
+
+abstract class CustomPolyDataCompanion[D[_]](
+  implicit instances: MacroInstances[CustomImplicits.type, MongoPolyAdtInstances[D]],
+) extends AbstractMongoPolyDataCompanion[CustomImplicits.type, D](CustomImplicits)
+
+/**
+  * This class tests (through its compilation) if implicit resolution conflicts that were
```
ddworak, line 19: you might want to add a link to PR / issue

ddworak (approved):
LGTM. Sorry for taking so long, I must've missed the email.

PR #435, opened 2023-03-18
Use the plainsbt plugin for build definition

ghik:
https://github.com/ghik/plainsbt


project/Commons.scala
```diff
@@ -17,18 +18,7 @@ import xerial.sbt.Sonatype.autoImport.sonatypeProfileName
 
 import scala.language.experimental.macros
```
ddworak, line 19: No longer needed :heart:

ddworak (approved):

PR #439, opened 2023-05-08
Redis client connection/reconnection fix

Kudyyy:
If PR fixes few issues:
- When client is reconnecting multiple times in a very short period of time it is possible that _Connected_ message from previous connection attempt will be handled after receiving _ConnectionClosed_. This was w race condition that causes client to think that it is connected
- _ConnectionClosed_ was ignored in _connecting_ state
- When Redis Server did not respond to our initial request, client was hung, waiting for the response without any timeout
- Retry strategy wasn't passed correctly in few places and there were some cases when client was trying to reconnect without any back-off
- Fix cluster initialisation procedure for tests

### DIFF
--- a/redis/src/main/scala/com/avsystem/commons/redis/actor/RedisConnectionActor.scala
+++ b/redis/src/main/scala/com/avsystem/commons/redis/actor/RedisConnectionActor.scala
@@ -18,6 +18,7 @@ import java.nio.{Buffer, ByteBuffer}
 import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.concurrent.duration.Duration
+import scala.util.Random
 
 final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
   extends Actor with ActorLazyLogging { actor =>
@@ -35,6 +36,7 @@ final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
     }
   }
 
+  private val random = new Random
   private var mustInitiallyConnect: Boolean = _
   private var initPromise: Promise[Unit] = _
 
@@ -42,6 +44,7 @@ final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
   private var incarnation = 0
   private var reservedBy = Opt.empty[ActorRef]
   private var reservationIncarnation = Opt.empty[Int]
+  private var currentConnectionId = Opt.empty[Int]
 
   private val queuedToReserve = new JArrayDeque[QueuedPacks]
   private val queuedToWrite = new JArrayDeque[QueuedPacks]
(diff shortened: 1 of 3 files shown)


redis/src/main/scala/com/avsystem/commons/redis/config/config.scala
```diff
@@ -169,6 +170,7 @@ case class ConnectionConfig(
   connectTimeout: OptArg[FiniteDuration] = OptArg.Empty,
   idleTimeout: OptArg[FiniteDuration] = OptArg.Empty,
   maxWriteSizeHint: OptArg[Int] = 50000,
+  initTimeout: FiniteDuration = 5.seconds,
```
Pinioo, line 173: How about `initResponseTimeout`, it would explain this setting better

Also, I think 5 seconds default value is too strict for a library, I would vote for at least 15.

Pinioo (commented):
Nice fix, GJ!

redis/src/main/scala/com/avsystem/commons/redis/actor/RedisConnectionActor.scala
```diff
@@ -92,6 +92,18 @@ final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
     }
   }
 
+  /**
+    * Without separate "reconnecting" state we would not be able to drop old messages
+    * (especially Connected message if arrived after ConnectionClosed because of some race conditions).
+    * Without this state, we would process old ones instead of dropping them and creating a new connection.
+    */
+  private def reconnecting(retryStrategy: RetryStrategy): Receive = {
```
Pinioo, line 100: I guess this behaviour should also react to other steering messages like Open, IncomingPacks, Release or Close

Kudyyy: My intention was to do not react to those as if we reconnecting, it means we already closed/lost connection so all old messages can be dropped. Handling those messages would cause reacting to messages from connection that does not longer exist

redis/src/main/scala/com/avsystem/commons/redis/actor/RedisConnectionActor.scala
```diff
@@ -107,9 +110,13 @@ final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
       //TODO: use dedicated retry strategy for initialization instead of reconnection strategy
       new ConnectedTo(connection, localAddress, remoteAddress).initialize(config.reconnectionStrategy)
```
Pinioo, line 111: I think we should also apply connectionId to `Connected`, we shouldn't go into initialized if `Connected` comes from different connection

redis/src/main/scala/com/avsystem/commons/redis/actor/RedisConnectionActor.scala
```diff
@@ -102,14 +105,23 @@ final class RedisConnectionActor(address: NodeAddress, config: ConnectionConfig)
...
-      readInitSender.foreach(_ ! ReadAck)
-    case _: ConnectionFailed =>
-      log.error(s"Connection attempt to Redis at $address failed")
-      tryReconnect(retryStrategy, new ConnectionFailedException(address))
+    case Connected(connection, remoteAddress, localAddress, connectionId) =>
+      if (currentConnectionId.forall(_ != connectionId)) {
+        log.error(s"Received Connected for connection different than currently trying to establish")
+        connection ! CloseConnection(immediate = true)
+      } else {
+        //TODO: use dedicated retry strategy for initialization instead of reconnection strategy
```
Pinioo, line 113: You miss `log.debug(s"Connected to Redis at $address")`, I guess it's an accident

PR #441, opened 2023-06-14
Fix nested arrayFilters in Mongo driver

### LINKED ISSUE #440: Invalid condition generation in filtered positional operator
Using the exact example from docs:
https://www.mongodb.com/docs/manual/reference/operator/update/positional-filtered/#update-all-array-elements-that-match-arrayfilters

 ```
case class Grades(id: Grades.ID, grades: List[Grade]) extends MongoEntity[Int]
object Grades extends MongoEntityCompanion[Grades]

case class Grade(grade: Int, mean: Int, std: Int)
object Grade extends MongoDataCompanion[Grade]

val update = Grades.ref(_.grades).updateFiltered(
  filter = _.ref(_.grade).gte(85),
  update = _.ref(_.mean).set(100)
)

```

We can see that the generated array query looks as follow:

```
println(update.toBsonAndArrayFilters._2.asScala.head.toJson)
> {"filter0": {"grade": {"$gte": 85}}}
```
instead of  `{"filter0.grade" : {"$gte": 85}}}`

This actually matches on documents in the array that are equal to `{"grade": {"$gte": 85}}`, instead of evaluating the condition on its fields, so it makes it impossible to actually query the documents in array.

### PR DESCRIPTION
**Pull request by sebaciv**
Fixes #440

### CODE REVIEW

File: mongo/jvm/src/main/scala/com/avsystem/commons/mongo/typed/MongoUpdate.scala [hunk 1]
```diff
@@ -45,9 +45,15 @@ sealed trait MongoUpdate[T] {
         case ArrayElementsQualifier.FirstMatching() => "$"
         case ArrayElementsQualifier.Each() => "$[]"
         case ArrayElementsQualifier.Filtered(filter) =>
-          val name = s"filter${arrayFilters.size}"
-          arrayFilters.add(Bson.document(name, filter.toBson))
-          s"$$[$name]"
+          val identifier = s"filter${arrayFilters.size}"
+          val (name, query) = filter match {
+            case MongoFilter.PropertyValueFilter(prop, filter) =>
```
ddworak, line 50: nit: `filter` is overloaded here with a different reference, not the most readable / error-prone

ddworak (approved):
LGTM with nit

PR #445, opened 2023-06-16
Scala update

sebaciv:
Scala 2.13.11, 2.11.18

### DIFF
--- a/core/src/test/scala/com/avsystem/commons/misc/SharedExtensionsTest.scala
+++ b/core/src/test/scala/com/avsystem/commons/misc/SharedExtensionsTest.scala
@@ -154,7 +154,12 @@ class SharedExtensionsTest extends AnyFunSuite with Matchers {
   }
 
   test("sourceCode") {
-    assert(123.sourceCode == "123")
+    // for some magical reason does not work on 'Int' type in Scala 2.13.11:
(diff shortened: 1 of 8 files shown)


core/src/test/scala/com/avsystem/commons/misc/SharedExtensionsTest.scala
```diff
@@ -154,7 +154,12 @@ class SharedExtensionsTest extends AnyFunSuite with Matchers {
   }
 
   test("sourceCode") {
-    assert(123.sourceCode == "123")
+    // for some magical reason does not work on 'Int' type in Scala 2.13.11:
+    // java.lang.NullPointerException: Cannot invoke "scala.reflect.internal.Symbols$Symbol.owner()" because the return value of "scala.reflect.internal.Trees$Tree.symbol()" is null
+    // scala.tools.nsc.typechecker.Typers$Typer.checkDubiousAdaptation$1(Typers.scala:5330)
+    // assert(123.sourceCode == "123")
```
sebaciv, line 160: Super strange compiler crash, seems like it cannot invoke macro that is an extension method on Int type? Not sure what to do about it.

ddworak: Issue in scala/bug to be added later on. Little chance we're getting the source code of a primitive anywhere.

ddworak (approved):

PR #448, opened 2023-07-11
Updates

ddworak:


project/Commons.scala
```diff
@@ -387,6 +382,7 @@ object Commons extends ProjectGroup("commons") {
     .settings(
       jvmCommonSettings,
       noPublishSettings,
+      crossScalaVersions := crossScalaVersions.value.take(1),
```
sebaciv, line 385: Why this change is needed?

ddworak: For `commons-benchmark` it's actually not needed at all, nice catch.
For `commons-benchmark-js` scalajs-benchmark is not available for 2.12 anymore and any commands depending on it will fail. Actually changing this doesn't prevent `sbt +commons-benchmark-js/compile` from failing anyway, will revert.

sebaciv (approved):

PR #482, opened 2023-08-31
Drop Scala 2.12 support

ddworak:


sebaciv:
There is a reference to 2.12 in README (https://maven-badges.herokuapp.com/maven-central/com.avsystem.commons/commons-core_2.12/badge.svg), but I'm not sure if it is relevant

ddworak:
Good call, also fixed SJS benchmarks which were broken for a while it seems.

PR #520, opened 2023-12-05
Create .scala-steward.conf and pin logback version

anetaporebska:


anetaporebska:
Irrelevant since logback is not present :+1:

PR #523, opened 2024-01-10
Redis 7

ddworak:


ddworak:
After verification it seems that initial compatibility issues with Redis 7 are now fixed on the server side and everything works as before. Moving on to testing on Redis 7.

PR #532, opened 2024-01-17
Add JavaApi annotation

sebaciv:
Add JavaApi annotation


core/src/main/scala/com/avsystem/commons/jiop/JavaApi.scala
```diff
@@ -0,0 +1,7 @@
+package com.avsystem.commons
+package jiop
+
+/**
+  * Marks Scala methods which exists only for Java compatibility / interoperability
+  */
+class JavaApi extends StaticAnnotation
```
ddworak, line 7: Let's move this to the JVM-specific module if possible.

core/src/main/scala/com/avsystem/commons/jiop/JavaApi.scala
```diff
@@ -0,0 +1,7 @@
+package com.avsystem.commons
+package jiop
+
+/**
+  * Marks Scala methods which exists only for Java compatibility / interoperability
```
ddworak, line 5: `method` or `exist`

sebaciv: Done

ddworak (approved):

PR #542, opened 2024-02-22
Reactive Publisher to Monix extensions for typed Mongo

sebaciv:
Extract conversions from `TypedMongoUtils` as proper extension methods, so they can be reused in other places

### DIFF
--- a/mongo/jvm/src/main/scala/com/avsystem/commons/mongo/reactive/ReactiveMongoExtensions.scala
+++ b/mongo/jvm/src/main/scala/com/avsystem/commons/mongo/reactive/ReactiveMongoExtensions.scala
@@ -1,0 +1,29 @@
+package com.avsystem.commons
+package mongo.reactive
+
+import monix.eval.Task
+import monix.reactive.Observable
+import org.reactivestreams.Publisher
+
+trait ReactiveMongoExtensions {
+  import ReactiveMongoExtensions._
+
+  implicit final def publisherOps[T](publisher: Publisher[T]): PublisherOps[T] = new PublisherOps(publisher)
+}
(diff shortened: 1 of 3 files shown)


mongo/jvm/src/main/scala/com/avsystem/commons/mongo/reactive/ReactiveMongoExtensions.scala
```diff
@@ -0,0 +1,29 @@
...
+    */
+  final class PublisherOps[T](private val publisher: Publisher[T]) extends AnyVal {
+    def asMonix: Observable[T] = Observable.fromReactivePublisher(publisher)
+
+    // prefer using the family of methods below for observables which are intended to only return a single document,
+    // mongo observable implementation sometimes logs an error on server-closed cursors
+    private def singleObservable: Task[Option[T]] = Task.fromReactivePublisher(publisher)
+
+    def headOptL: Task[Opt[T]] = singleObservable.map(_.toOpt)
+    def headOptionL: Task[Option[T]] = singleObservable.map(_.filterNot(_ == null))
```
ddworak, line 25: Can `singleObservable` contain `Some(null)` in the wild? :scream:

If that's the case, maybe we should remove `singleObservable` and use this as a starting point for other methods tbh.

sebaciv: monix.eval.internal.TaskConversions#fromReactivePublisher does not protect against it (`Some(a)`), so if Mongo driver returns `null` we will end up with `Some(null)` here

Previous version of `TypedMongoUtils` had `null` protection as well

ddworak (approved):

PR #560, opened 2024-04-09
Add `Opt.when` method

halotukozak:
`Opt.when` is similar to `optIf`, but takes `Opt[?]` parameter. The inspiration came from `Task.when`.

It may simplify code like this:

`if (predicate(sth)) sthElse.orElse(sthMoreElse.map(_.sth)) else Opt.Empty`

to

`Opt.when(predicate(sth)){ sthElse.orElse(sthMoreElse.map(_.sth)) }`

### DIFF
--- a/core/src/main/scala/com/avsystem/commons/misc/Opt.scala
+++ b/core/src/main/scala/com/avsystem/commons/misc/Opt.scala
@@ -28,6 +28,20 @@ object Opt {
     def foreach[U](f: A => U): Unit = self filter p foreach f
     def withFilter(q: A => Boolean): WithFilter[A] = new WithFilter[A](self, x => p(x) && q(x))
   }
+
+  final class LazyOptOps[A](private val opt: () => Opt[A]) extends AnyVal {
+    /** When a given condition is true, evaluates the `opt` argument and returns it.
+      * When the condition is false, `opt` is not evaluated and `Opt.Empty` is
+      * returned.
+      */
+    def when(cond: Boolean): Opt[A] = if (cond) opt() else Opt.Empty
+    /** Unless a given condition is true, this will evaluate the `opt` argument and
+      * return it. Otherwise, `opt` is not evaluated and `Opt.Empty` is returned.
+      */
+    @inline def unless(cond: Boolean): Opt[A] = when(!cond)
+  }
+
+  implicit def lazyOptOps[A](opt: => Opt[A]): LazyOptOps[A] = new LazyOptOps(() => opt)
 }
 
 /**
@@ -41,7 +55,8 @@ object Opt {
   * Please be aware of that tradeoff.
   */
 final class Opt[+A] private(private val rawValue: Any) extends AnyVal with OptBase[A] with Serializable {
-  import Opt._
+
+  import Opt.*
 
   private def value: A = rawValue.asInstanceOf[A]
 
@@ -140,7 +155,7 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with OptBa
     if (isEmpty) Iterator.empty else Iterator.single(value)
 
   @inline def toList: List[A] =
-    if (isEmpty) List() else new ::(value, Nil)
+    if (isEmpty) List() else value :: Nil
 
   @inline def toRight[X](left: => X): Either[X, A] =
     if (isEmpty) Left(left) else Right(value)
--- a/core/src/test/scala/com/avsystem/commons/misc/OptTest.scala
+++ b/core/src/test/scala/com/avsystem/commons/misc/OptTest.scala
@@ -66,4 +66,16 @@ class OptTest extends AnyFunSuite {
     val seq: Seq[Int] = Opt(5).mapOr(Nil, i => 0 until i)
     assert(seq == (0 until 5))
   }
+
+  test("Opt.{when, unless}") {
+    val opt = Opt(42)
+
+    assert(opt.when(true) == opt)
+    assert(opt.when(false) == Opt.Empty)
+    assert(Opt(fail("Parameter should not be evaluated")).when(false) == Opt.Empty)
+
+    assert(opt.unless(false) == opt)
+    assert(opt.unless(true) == Opt.Empty)
+    assert(Opt(fail("Parameter should not be evaluated")).unless(true) == Opt.Empty)
+  }
 }


core/src/test/scala/com/avsystem/commons/misc/OptTest.scala
```diff
@@ -66,4 +66,10 @@ class OptTest extends AnyFunSuite {
     val seq: Seq[Int] = Opt(5).mapOr(Nil, i => 0 until i)
     assert(seq == (0 until 5))
   }
+
+  test("Opt.when") {
+    val opt = Opt(42)
+    assert(Opt.when(true)(opt) == opt)
+    assert(Opt.when(false)(opt) == Opt.Empty)
+  }
```
ddworak, line 74: the laziness in parameter evaluation is not tested

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -140,7 +143,7 @@ final class Opt[+A] private(private val rawValue: Any) extends AnyVal with OptBa
     if (isEmpty) Iterator.empty else Iterator.single(value)
 
   @inline def toList: List[A] =
-    if (isEmpty) List() else new ::(value, Nil)
+    if (isEmpty) List() else new::(value, Nil)
```
ddworak, line 146: nit: maybe just `else value :: Nil`

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -14,6 +14,8 @@ object Opt {
     if (value != null) new Opt[A](value)
     else throw new NullPointerException
 
+  def when[A](cond: Boolean)(value: => Opt[A]): Opt[A] = if (cond) value else Opt.Empty
```
ddworak, line 17: I know we're spotty with docs but you can be better. Let's reuse `scala.Option#when` docs (I'm actually suprised this was not your inspiration :D)

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -14,6 +14,8 @@ object Opt {
...
+
```
ddworak, line 18: If we're adding `when`, we should probably add `unless`.

ddworak (changes requested):

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
ddworak: Right, then I'm not sure it should be... WDYT @sebaciv?

sebaciv: `a: => A` use-case is covered by `.optIf` extension method, so this variant of `when` seems useful

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
sebaciv: ++

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
ddworak: I'm just not sure this is the right API. It's misleading given the similarities to Option and we always tried to make Opt a drop-in replacement as much as possible. Maybe we should just add a method / extension method instead.

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
halotukozak: intelliJ's formatting :/

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
halotukozak: I didn't know about the fact that `we always tried to make Opt a drop-in replacement as much as possible` If you prefer the extension method, I'd go for sth like the combination of `optIf` and `flatten`. It would be sth like a lazy filter, but with the `p: Boolean` predicate, not the `p: A: Boolean` one. So, again, it wouldn't be consistent.

halotukozak:
```scala
 implicit class OptOps[A](opt: => Opt[A]) {
    def lazyFilter(p: Boolean): Opt[A] = if (p) opt else Opt.Empty
  }


  Opt(5).lazyFilter(1 == 1) // Opt(5)
  Opt(5).lazyFilter(1 == 2) // Opt.Empty
  
```

ddworak: how about `.when(Boolean)` and `.unless(Boolean)` as extensions?

halotukozak: It is not in line with my sense of aesthetics, but it can be ¯\_(ツ)_/¯.

ddworak: I don't mind `lazyFilter`, it's just `lazyFilterNot` that makes me a bit worried :D

ddworak: It can. You just need to avoid implicit class sugar and code it up yourself, as in `com.avsystem.commons.SharedExtensionsUtils.LazyUniversalOps`

core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -188,6 +189,18 @@ object SharedExtensionsUtils extends SharedExtensions {
       }
   }
 
+  implicit class LazyOptOps[A](private val opt: () => Opt[A]) extends AnyVal {
```
ddworak, line 192: `implicit` no longer needed when you have a manual implicit conversion. It might even generate a conflict with unfortunate enough imports. Also make it `final` for additional performance optimizations by the JVM.

core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -26,6 +25,8 @@ trait SharedExtensions {
 
   implicit def futureCompanionOps(fut: Future.type): FutureCompanionOps.type = FutureCompanionOps
 
+  implicit def lazyOptOps[A](opt: => Opt[A]): LazyOptOps[A] = new LazyOptOps(() => opt)
```
ddworak, line 28: You don't need this in `SharedExtensions` since `Opt` is a type we control. The compiler will find the implicit if it's in `Opt` companion just as well.

ddworak (changes requested):

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
```diff
@@ -28,6 +28,20 @@ object Opt {
     def foreach[U](f: A => U): Unit = self filter p foreach f
     def withFilter(q: A => Boolean): WithFilter[A] = new WithFilter[A](self, x => p(x) && q(x))
   }
+
+  class LazyOptOps[A](private val opt: () => Opt[A]) extends AnyVal {
```
ddworak, line 32: nit: final

ddworak (approved):
LGTM after nits.

@sebaciv do you want to take a final look?

sebaciv (approved):

core/src/main/scala/com/avsystem/commons/misc/Opt.scala
Kurtzz: Imho, it is useless and misleading API.

It can be only used in the same place the Opt is created:
```scala
Opt(value).when(condition)
```
or maybe in a method, if `Opt` is passed as by-name argument or using `Function0[Opt[_]]`.
It brings nothing when it is assigned to a val, cause `.filter` can be used at this point.

`.optIf(condition).flatten` does the same and it is not much longer expression then `.when(condition)`

Besides, need of such an API, imho, means that the issue is in the implementation. Probably it should be `Opt[() => T]` or `Opt[SomeEvaluator]` or just `SomeEvaluator` that can be `optIf`ed.

PR #566, opened 2024-04-16
Revert "Add `Opt.when` method"

halotukozak:
Reverts AVSystem/scala-commons#560

### DIFF
--- a/core/src/main/scala/com/avsystem/commons/misc/Opt.scala
+++ b/core/src/main/scala/com/avsystem/commons/misc/Opt.scala
@@ -28,20 +28,6 @@ object Opt {
     def foreach[U](f: A => U): Unit = self filter p foreach f
     def withFilter(q: A => Boolean): WithFilter[A] = new WithFilter[A](self, x => p(x) && q(x))
   }
-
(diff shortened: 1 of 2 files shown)


halotukozak:
The proposed addition of the methods to the Opt class has raised several concerns among the team. Firstly, there's uncertainty about its necessity and usefulness within the codebase. The method's purpose, which appears to be similar to existing methods like `Task.when` and `Option.when`, is questioned, as it introduces redundancy without clear benefits.

Moreover, there are doubts about the clarity and readability of the proposed method. The names may not accurately convey its functionality, potentially leading to confusion for developers working with the code. Additionally, concerns are raised about the method's impact on code execution efficiency, particularly regarding lazy evaluation and performance overhead.

Overall, the consensus emerges that the proposed addition does not address a significant problem or offer substantial improvements to the codebase. Therefore, there's a sentiment that investing time and effort into implementing and maintaining this feature may not be justified, leading to the decision to revert the change.

ddworak:
I'm okay (just as I was when accepting the original PR) with the current shape of this rather innocent extension method for `Opt`. Unless an actual improvement proposal is submitted, I don't see any benefit in merging this.

Issue #500, opened 2023-10-12
Handle thrown exception outside the monadic context of Task


halotukozak:
`Task(sth).map(throw Error())` doesn't lead to failing the Task, because `throw Error()` has `Nothing` type, so it has also lambda type. If such a check is possible, it should be included in `commons-analyzer`.

halotukozak:
@ddworak can we either proceed this issue or close it?

ddworak:
Let's proceed, sorry for missing your changes after initial review.

PR #585, opened 2024-06-12
add NothingAsFunctionArgument rule to the Analyzer Plugin

halotukozak:


halotukozak:
Closes #500

halotukozak:
It's a draft. We can consider wider solution to warn when `Nothing` provided in place of Function/PartialFunction argument.

analyzer/src/main/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInMonixScope.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    case Apply(TypeApply(Select(prefix: Tree, TermName("map")), List(_)), List(Throw(_))) if prefix.tpe <:< monixTaskTpe =>
+      report(tree.pos, "exception thrown not in Monix Task scope ")
+
+    case tree =>
+      tree.children.foreach(checkDiscardedNothing(_, discarded = false))
+  }
+
+  def analyze(unit: CompilationUnit): Unit = checkDiscardedNothing(unit.body, discarded = false)
+}
```
ddworak, line 59: I'm not intimately familiar with analyzer internals, but I hope they are not complicated enough to force the programmer to match every construct Scala has to offer. I did a small check and your tests pass with this implementation:

```scala
final class ThrownExceptionNotInMonixScope(g: Global) extends AnalyzerRule(g, "thrownExceptionNotInMonixScope") {

  import global.*

  lazy val monixTaskTpe: Type = classType("monix.eval.Task") match {
    case NoType => NoType
    case tpe => TypeRef(NoPrefix, tpe.typeSymbol, List(definitions.AnyTpe))
  }

  def analyze(unit: CompilationUnit): Unit = {
    unit.body.foreach(analyzeTree {
      case t@Apply(TypeApply(Select(prefix: Tree, TermName("map")), List(_)), List(Throw(_))) if prefix.tpe <:< monixTaskTpe =>
        report(t.pos, "exception thrown not in Monix Task scope ")
    })
  }
}
```

analyzer/src/main/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInMonixScope.scala
```diff
@@ -0,0 +1,59 @@
...
+    case Match(_: Tree, cases: List[CaseDef]) =>
+      cases.foreach(checkDiscardedNothing(_, discarded))
+
+    case Annotated(_: Tree, arg: Tree) =>
+      checkDiscardedNothing(arg, discarded)
+
+    case Typed(expr: Tree, _: Tree) =>
+      checkDiscardedNothing(expr, discarded)
+
+    case Apply(TypeApply(Select(prefix: Tree, TermName("map")), List(_)), List(Throw(_))) if prefix.tpe <:< monixTaskTpe =>
```
ddworak, line 51: Should we only handle mapping or consider other methods as well?

ddworak (changes requested):

analyzer/src/main/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInMonixScope.scala
halotukozak: tbh I was inspired by `DiscardedMonixTask`

analyzer/src/main/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInMonixScope.scala
halotukozak: I am gonna ask: should we also consider constructors?

ddworak: I don't understand the question: `Task(throw new RuntimeException("test"))` (not a constructor, but a factory) would work correctly in this case. 

In general, I'm not sure we should cover too many classes - let's start small - but I was thinking about the ones most closely related to the issue you were originally testing (e.g. `Task#flatMap`).

halotukozak: I'd like to extend the scope, because in general we do not want to do `throw Ex` in place of possible `_ => throw Ex`, don't we? And yes, `Task.apply` is a function, but `new Sth(...)` is not and also we can cover this case, imo.

halotukozak:
@ddworak Can we either proceed this MR or close it?

analyzer/src/main/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInFunction.scala
```diff
@@ -0,0 +1,18 @@
+package com.avsystem.commons
+package analyzer
+
+import scala.tools.nsc.Global
+
+final class ThrownExceptionNotInFunction(g: Global) extends AnalyzerRule(g, "thrownExceptionNotInFunction") {
```
ddworak, line 6: The name `ThrownExceptionNotInFunction` suggests it checks for exceptions thrown outside functions, but it’s more about misusing thrown exceptions as function arguments.

Good rule names tend to:
- Describe the smell, not the fix, so users scan a list of rules to see what they catch.
- Include the syntactic location (“…Argument”, “…Literal”, “…Parameter”). This avoids ambiguity with similar rules.
- Stay concise
- Use positive form if possible - easier to skim than “Not”/“Missing”

Therefore, I'd propose sth like `ThrowAsFunctionArgument`. It's precise (the problem is exactly a throw used as an argument where a function is expected) and reads naturally.

analyzer/src/test/scala/com/avsystem/commons/analyzer/ThrownExceptionNotInFunctionTest.scala
```diff
@@ -0,0 +1,209 @@
...
+           |  do sth.$function(_ => throw ex) while (true)
+           |}
+           |
+           |""".stripMargin
+      )
+    }
+  }
+
+  test("Testing multiple arguments") {
+    assertErrors(26,
```
ddworak, line 102: Debugging why an `assertErrors(26, ...` failed is impossible without editing the tests. Let's work on the developer experience in this test by using sth like `org.scalatest.Inspectors` in conjunction with `assertErrors(1` and `assertNoErrors` (example, I'm sure you can suggest the best solution available).

ddworak (changes requested):

PR #606, opened 2024-07-18
Fix Redis tests by with new, valid certificate

halotukozak:


sebaciv (approved):

ddworak:
@halotukozak why do we need this so quickly after https://github.com/AVSystem/scala-commons/pull/594? The certificates added there had a long expiration date.

halotukozak:
> @halotukozak why do we need this so quickly after #594? The certificates added there had a long expiration date.

I have no idea ¯\_(ツ)_/¯. I've found the solution, but not the explanation.

ddworak:
Please file an issue with all the details as the tests were passing before your change and it's not obvious. We will have to look into it to prevent it from reoccurring.

halotukozak:
> Please file an issue with all the details as the tests were passing before your change and it's not obvious. We will have to look into it to prevent it from reoccurring.

Done #607

PR #609, opened 2024-07-26
Add .tapFailure to TryOps

najder-k:
- works basically like `tapError` on Task
- used to avoid doing `failed.foreach` on task, which creates an unnecessary exception on Try.Success


core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -461,6 +461,19 @@ object SharedExtensionsUtils extends SharedExtensions {
...
+    /**
+      * Apply side-effect only if Try is a failure.
+      *
+      * Don't use .failed projection here, because it unnecessarily creates Exception in case of Success,
+      * which is an expensive operation.
+      */
+    def tapFailure(action: Throwable => Unit): Try[A] = tr match {
+      case Success(_) => tr
+      case Failure(throwable) =>
+        action(throwable)
```
ddworak, line 474: What do you want to do when `action(throwable)` throws an exception? Not sure if current choice would be my default, but I'd like to hear your thoughts first. Please document your choice as well.

ddworak (changes requested):

core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
ddworak: My thoughts were that any `.tap` method should never alter the result. The libraries are also on the fence: `Task#tapError` handles, `Flow#wireTap` I think too but `Iterator#tapEach` and `Observable#doOnNext` do not. I'd probably replicate `Task` approach and use the same name, but I'm okay with other behaviour if it's documented and tested.

halotukozak: fixed with catching nonfatal exception. Tests added. Can we proceed this MR?

najder-k: You actually went for the 'tap should never alter the result' way, but tbh I'm fine with it too (after some time passed, probably even more than with the `Task#tapError` approach)

Since it's more your MR than mine at this point, I'll take over the review :D

core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -463,9 +463,13 @@ object SharedExtensionsUtils extends SharedExtensions {
       if (tr.isFailure) OptArg.Empty else OptArg(tr.get)
 
     /**
-      * Apply side-effect only if Try is a failure.
+      * Apply side-effect only if Try is a failure. The provided `action` function will be called with the
+      * throwable from the failure case, allowing you to perform operations like logging or error handling.
+      * 
+      * Any exceptions thrown by the `action` function are caught and ignored, ensuring that this method
```
ddworak, line 469: This is not true. You've literally just added a test that shows it's not all exceptions.

ddworak (changes requested):

ddworak (approved):

PR #611, opened 2024-08-02
Implement JettyRPCFramework on ee10 for Jetty 12

ddworak:


jetty/src/main/scala/com/avsystem/commons/jetty/rpc/JettyRPCFramework.scala
```diff
@@ -89,30 +88,38 @@ object JettyRPCFramework extends StandardRPCFramework with LazyLogging {
...
+          completeWith(Future.fromTry(call).flatMapNow(handlePost).andThenNow {
             case Success(responseContent) =>
               response.setContentType(MimeTypes.Type.APPLICATION_JSON.asString())
               response.setCharacterEncoding(StandardCharsets.UTF_8.name())
               response.getWriter.write(responseContent.s)
             case Failure(t) =>
               response.sendError(HttpStatus.INTERNAL_SERVER_ERROR_500, t.getMessage)
               logger.error("Failed to handle RPC call", t)
-          }.andThenNow { case _ => async.complete() }
+          })
```
sebaciv, line 120: That does not seem right to me, `asyncContext.complete()` is not called after `Future`. Argument `code` in `completeWith` method is a Future, so:
```
code // some future running in background
asyncContext.complete() // imidietly closes context without waiting for actual completion

```

sebaciv, line 120: That does not seem right to me, `asyncContext.complete()` is not called after `Future` finishes. Argument `code` in `completeWith` method is a Future, so:
```
def completeWith(code: => Unit): Unit = {
  code // some future running in background
  asyncContext.complete() // immediately closes context without waiting for actual completion
}
```

sebaciv (changes requested):

jetty/src/main/scala/com/avsystem/commons/jetty/rpc/JettyRPCFramework.scala
```diff
@@ -1,21 +1,20 @@
...
+import org.eclipse.jetty.ee10.servlet.ServletContextHandler
 import org.eclipse.jetty.http.{HttpMethod, HttpStatus, MimeTypes}
-import org.eclipse.jetty.server.handler.AbstractHandler
-import org.eclipse.jetty.server.{Handler, Request}
+import org.eclipse.jetty.server.Handler
 
-import scala.concurrent.duration._
+import java.nio.charset.StandardCharsets
+import java.util.concurrent.atomic.AtomicBoolean
+import scala.concurrent.duration.*
```
sebaciv, line 16: Nit: switch to Scala 3 syntax? IMO we should do it in all classes or keep using Scala 2 _

jetty/src/main/scala/com/avsystem/commons/jetty/rpc/JettyRPCFramework.scala
ddworak: Yup, copied that from udash-rest without checking yet. Should be better now if I understand the intention correctly.

sebaciv (approved):

PR #623, opened 2024-08-22
add byNameImplicitParameter rule

halotukozak:


ddworak:
I do agree that this construct is mixing two complex and powerful Scala concepts, however I don't think the feature is "globally wrong" to warrant such rule. Let's strive to use the principle of least power when designing and reviewing the code.

PR #655, opened 2024-12-06
introduce ImplicitOptArg

halotukozak:
`ImplicitOptArg` is like `OptArg`, but also accept implicit parameters (useful for DI or context parameters)

Example usage:
```
  class ClientSession

  val NoSession = ImplicitOptArg.Empty[ClientSession]

  class TransactionManager {
    def execute(transaction: ClientSession => Unit): Unit = ???
  }

  val transactionManager = new TransactionManager

  def doSomething(implicit session: ImplicitOptArg[ClientSession]): Unit = ???

  doSomething(using NoSession) // explicitly passing NoSession

  transactionManager.execute { implicit session =>
    doSomething // implicit session is passed
  }
```

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,23 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
-    <JetCodeStyleSettings>
-      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
-        <value>
-          <package name="java.util" alias="false" withSubpackages="false" />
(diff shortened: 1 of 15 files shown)


halotukozak:
> Thank you for proposing this idea. I don't think we should encourage implicit parameters with default values, especially when combined with optionality. `OptArg`s are often used in contexts where explicitness is highly desired (e.g. configuration, application bootstrap).

we do not `encourage implicit parameters with default values`. The default ones must be passed explicitly (how it's already done internally with MaybeClientSession in DHCP and CAE with UserContext).
But I see, this mechanism could be abused

PR #662, opened 2024-12-24
Deprecate commons-redis

ddworak:
Most of our internal products migrated or are in the process of migrating to Redisson so I propose to release a deprecation. 

This module is particularly flaky and maintenance intensive so I've also removed it from CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@
     [`OptRef`](http://avsystem.github.io/scala-commons/api/com/avsystem/commons/misc/OptRef.html) (implemented as value
     classes)
   * [Components](docs/Components.md) and Dependency Injection library
-* `commons-redis` - [Scala driver for Redis](docs/RedisDriver.md)
 * `commons-macros` contains implementations of macros used in other modules and reusable macro utilities:
   * `MacroCommons` trait with several convenience functions for implementing macros
   * `TypeClassDerivation` - implements infrastructure for automatic type class derivation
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -33,7 +33,6 @@
     [`OptRef`](http://avsystem.github.io/scala-commons/api/com/avsystem/commons/misc/OptRef.html) (implemented as value
     classes)
   * [Components](docs/Components.md) and Dependency Injection library
-* `commons-redis` - [Scala driver for Redis](docs/RedisDriver.md)
```
mikegpl, line 36: suggestion: If we're keeping the Redis docs (I noticed you added deprecation notice in `RedisDriver.md`) until we remove the driver code entirely, IMO we should keep the entrypoint to its docs too.

redis/src/main/scala/com/avsystem/commons/redis/RedisClusterClient.scala
```diff
@@ -45,6 +45,7 @@ import scala.concurrent.duration._
   * @param config    client configuration - [[ClusterConfig]]
   * @param clusterStateObserver optional observer for monitoring client's state and connections - [[ClusterStateObserver]]
   */
+@deprecated("Redis driver is scheduled for removal. Use a different library, e.g. redisson.", "2.21.0")
```
mikegpl, line 48: suggestion: Extend this deprecation notice (and other ones as well) with info that this is no longer tested since `2.21.0`. If, for some maintenance reasons, we release new version in the future and someone encounters unexpected behavior, it might be a good clue for searching for the root cause.

mowczare:
LGTM

mowczare (approved):

mikegpl (approved):

PR #674, opened 2025-02-13
Utilities for extracting GenCodec field / type names

sebaciv:
Utilities for extracting GenCodec field / type names

### DIFF
--- a/core/src/test/scala/com/avsystem/commons/serialization/GenCodecUtilsTest.scala
+++ b/core/src/test/scala/com/avsystem/commons/serialization/GenCodecUtilsTest.scala
@@ -1,0 +1,63 @@
+package com.avsystem.commons
+package serialization
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class GenCodecUtilsTest extends AnyFunSuite with Matchers {
+  import GenCodecUtilsTest.*
+
+  test("plain case class") {
+    val name = GenCodecUtils.codecTypeName[Foo]
+    name shouldBe "Foo"
+  }
+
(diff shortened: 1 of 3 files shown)


core/src/test/scala/com/avsystem/commons/serialization/GenCodecUtilsTest.scala
```diff
@@ -0,0 +1,63 @@
+package com.avsystem.commons
+package serialization
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class GenCodecUtilsTest extends AnyFunSuite with Matchers {
+  import GenCodecUtilsTest.*
+
```
ddworak, line 9: If we change behaviour of GenCodec, but not the behaviour of this class, we may miss the discrepancy. Maybe we should write a custom output that will make it simple to compare if the name is also equal to what gencodec actually outputs.

macros/src/main/scala/com/avsystem/commons/macros/serialization/GenCodecUtilMacros.scala
```diff
@@ -0,0 +1,47 @@
+package com.avsystem.commons
+package macros.serialization
+
+import scala.annotation.tailrec
+import scala.reflect.macros.blackbox
+
+class GenCodecUtilMacros(ctx: blackbox.Context) extends CodecMacroCommons(ctx) {
```
ddworak, line 7: Do you think this macro could share some code with GenCodec macros for more intended coupling now that they are in the same module?

ddworak (changes requested):

PR #696, opened 2025-04-17
Analyzer enhancements

halotukozak:


docs/Analyzer.md
```diff
@@ -31,6 +31,13 @@ Here's a list of currently supported rules:
...
 | `basePackage`              | warning       | Checks if all sources are within the specified base package                                                                                                                                          |
+| `catchThrowable`           | warning       | Makes sure that code does not catch `Throwable` directly, which can hide critical errors like `OutOfMemoryError`                                                                                     |
+| `finalValueClasses`        | warning       | Makes sure that value classes are marked final.                                                                                                                                                      |
+| `finalCaseClasses`         | warning       | Makes sure that case classes are marked final. Does not affect inner classes.                                                                                                                        |
```
najder-k, line 36: add context - does not affect inner classes due to SI-4440

analyzer/src/main/scala/com/avsystem/commons/analyzer/ImplicitValueClasses.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  import global.*
+  import definitions.*
+
+  private lazy val defaultClasses = Set[Symbol](AnyClass, AnyValClass, ObjectClass)
+
+  private lazy val reportOnNestedClasses = argument match {
+    case null => false
+    case a => a.toBoolean
+  }
```
najder-k, line 16: not a fan of the non-descriptive 'true' parameter, some mode enum like `all`, `top-level` would be better (with a default of course)

analyzer/src/test/scala/com/avsystem/commons/analyzer/ImplicitValueClassesTest.scala
```diff
@@ -0,0 +1,230 @@
...
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class ImplicitValueClassesSuite extends AnyFunSuite with AnalyzerTest {
+  test("implicit final class extending AnyVal should pass") {
+    assertNoErrors(
+      //language=Scala
+      """
+        |object whatever {
+        |  implicit final class GoodImplicitClass(val x: Int) extends AnyVal {
```
najder-k, line 12: NIT:

not a fan of implicit classes which have a `val` on their wrapped type - it should be `private val` basically always, otherwise you can do something like this:

```scala
42.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x
```

consider integrating it into this rule (since AnyVals need `val`, but it can be a `private val`)

najder-k (changes requested):
Heck yeah brother, solid contrib 💪  just a few minor changes necessary

docs/Analyzer.md
```diff
@@ -16,21 +16,27 @@ and `@nowarn` annotation for warning suppression.
...
+| `finalValueClasses`        | warning       | Makes sure that value classes are marked final.                                                                                                                                                                                   |
+| `finalCaseClasses`         | warning       | Makes sure that case classes are marked final. Does not affect inner classes due to SI-4440                                                                                                                                       |
+| `implicitParamDefaults`    | warning       | Makes sure that default values are not defined for implicit parameters                                                                                                                                                            |
+| `implicitValueClasses`     | warning       | Makes sure that implicit classes extend `AnyVal` (when applicable). Available options: `top-level-only` (default) - only checks implicit classes in top-level objects; `all` - checks both top-level and nested implicit classes. |
```
najder-k, line 38: NIT:
 it's not clear how to pass the options to this flag - try formatting them like https://docs.scala-lang.org/overviews/compiler-options/index.html

najder-k (approved):
Lgtm w/ one nit

PR #721, opened 2025-07-10
improve handling of custom handler CaseDefs in CatchThrowable

### LINKED ISSUE #720: CatchThrowable should not catch custom handlers
```scala
object CustomHandler {
  def apply[T](): PartialFunction[Throwable, T] = ???
}

try {} catch CustomHandler()
```

should be valid code. Instead, it reports a warning with no Position.

Note, that this code:
```scala
object CustomHandler {
  def apply[T](): PartialFunction[Throwable, T] = ???
}

try {} catch {
  case _: Throwable => ???
}

try {} catch CustomHandler()
```

generates almost the same `CaseDef`s

```scala
//case _: Throwable 
case (_: Throwable) => scala.Predef.???
//CustomHandler()
case (x$1 @ (_: Throwable)) => {
  <artifact> val catchExpr$1: PartialFunction[Throwable,Nothing] = CustomHandler.apply[Nothing]();
  if (catchExpr$1.isDefinedAt(x$1))
    catchExpr$1.apply(x$1)
  else
    throw x$1
}
```

halotukozak:
Closes #720


halotukozak:
Tested internally on CAE, works 🥳

sebaciv (approved):

PR #726, opened 2025-08-01
correct method anyMatch calls in ScalaJ*Stream classes

halotukozak:
10 years and no one noticed?


ddworak:
Hope it breaks something

ddworak (approved):

PR #752, opened 2025-09-24
add lifecycle components

halotukozak:
### Summary of Changes

- Refactored lifecycle components to enhance initialization and disposal handling.
- Introduced new lifecycle component traits and corresponding tests to ensure reliability. 

### Checklist

- [ ] Tests have been added or updated (if applicable).
- [ ] Documentation has been updated (if applicable).

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -164,6 +164,69 @@ object MyApp extends Components {
 }
 ```
 
+## Lifecycle-aware components
+
+Many application services have explicit initialization and shutdown phases. Components support this out of the box.
+
+There are four lifecycle traits you can implement on the values produced by your components:
+
+- InitializingComponent — synchronous init hook def init(): Unit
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/avsystem/commons/di/Component.scala
```diff
@@ -171,10 +171,15 @@ final class Component[+T](
...
+              .flatMap {
+                case component: AsyncInitializingComponent => component.init().map(_ => component)
+                case component: InitializingComponent => Future(component.init()).map(_ => component)
+                case component => Future.successful(component)
+              }
               .recoverNow {
                 case NonFatal(cause) =>
                   throw ComponentInitializationException(this, cause)
-              }
+              }.asInstanceOf[Future[T]]
```
ddworak, line 182: A simple `.flatMap { value => value match {` with mappings to `value` allows to avoid casting here.

PR #829, opened 2026-04-13
remove OrderingOps and some IteratorOps

halotukozak:
These extensions won't be taken either way. The compiler always picks the member methods

### DIFF
--- a/core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
+++ b/core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
@@ -3,8 +3,8 @@ package com.avsystem.commons
 import com.avsystem.commons.concurrent.RunNowEC
 import com.avsystem.commons.misc.*
 
-import scala.annotation.{nowarn, tailrec}
-import scala.collection.{mutable, AbstractIterator, BuildFrom, Factory}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/avsystem/commons/SharedExtensions.scala
```diff
@@ -53,6 +53,7 @@ trait SharedExtensions {
 
   implicit def iteratorCompanionOps(it: Iterator.type): IteratorCompanionOps.type = IteratorCompanionOps
 
+  @deprecated("Scala 2.13 has native implementation for these utils", "2.27.0")
```
ddworak: If they are native and were never picked up, why do we deprecate and not remove?

halotukozak: binary compat. but... we don't publish scala 2.12.x artifacts, so it's hard (but possible!) to get MethodNotFoundException

ddworak: we're gonna have to break soon anyway, but I guess we can stay with deprecated for a while

ddworak (approved):

PR #852, opened 2026-05-29
fix: badSingletonComponent false-positive on extension methods

### LINKED ISSUE #851: analyser: BadSingletonComponent rule incorrectly disallows extension methods to Component
## Problem

The `badSingletonComponent` analyzer rule false-positives when `singleton(...)` is followed by an extension method reached through an implicit conversion (e.g. an `implicit class` or an `implicit def` returning a wrapper).

The rule's `Unwrap` extractor peels trailing shape-preserving method chains so it can recognise the def body as rooted in `cached(...)`. It does so by requiring `prefix.tpe =:= t.tpe` on the chained `Select`. For an extension method, the implicit conversion intercedes — the prefix has the adapter's type (e.g. `ComponentExt[T]`), not `Component[T]` — so the peel is refused and the rule reports the body as malformed even though the chain is in fact shape-preserving.

## Minimal reproduction

```scala
import com.avsystem.commons.di._

object test extends Components {
  implicit class ComponentExt[T](c: Component[T]) {
    def chained(f: T => Unit): Component[T] = c
  }

  def good: Component[Int] = singleton(123)                    // accepted (correct)
  def withExt: Component[Int] = singleton(123).chained(_ => ()) // FALSE POSITIVE
}
```

`withExt` is rejected with:

> `[AVS] singleton(...) macro can only be used as a body of a parameterless method in a Components trait implementation`

## Expected

Both defs accepted. The chain is shape-preserving with respect to `Component[T]`, and `chained` is just a thin wrapper that returns the same component — caching and the dependency graph are unaffected.

## Impact

This blocks ergonomic Components extensions (e.g. a wrapper for `destroyWith` that would introduce exception handling). The rule has no per-call-site suppression mechanism at `*` (Info) level, and downgrading to Warn just to use `@nowarn` is invasive (one annotation per call site).

## Fix

https://github.com/AVSystem/scala-commons/pull/852

mikegpl:
Resolves #851 

The Unwrap extractor inside BadSingletonComponent peels trailing shape-preserving method chains (e.g. `singleton(x).dependsOn(y)`) so the rule can recognise the body as rooted in a `cached(...)` call. It does that by requiring `prefix.tpe =:= t.tpe` - i.e. the prefix of the trailing `.method(...)` must have the same type as the whole expression.

That check breaks for extension methods reached via an implicit conversion.

This commit adds one extra Unwrap case that peels a method call reached through a single-argument adapter Apply, gated by `orig.tpe =:= t.tpe` on the adapter's converted argument. The invariant the rule relies on ("the chain returns the same `Component[T]` type that the cached singleton produced") is preserved, and the new case does not relax the checks the original test cases rely on: bad usages were reported by arm 2 (`t.symbol == cachedSym` catch-all), which is independent of Unwrap.

Tests: extend BadSingletonComponentTest with `extension methods reached via implicit conversion are accepted`, covering the bare extension call, the extension-then-native chain, and the native-then-extension chain. The original `general` test (5 errors) is unaffected.

### DIFF
--- a/analyzer/src/main/scala/com/avsystem/commons/analyzer/BadSingletonComponent.scala
+++ b/analyzer/src/main/scala/com/avsystem/commons/analyzer/BadSingletonComponent.scala
@@ -25,6 +25,13 @@ class BadSingletonComponent(g: Global) extends AnalyzerRule(g, "badSingletonComp
       case Typed(expr, _) => unapply(expr)
       case Annotated(_, expr) => unapply(expr)
       case UnwrapApply(Select(prefix, _)) if prefix.tpe =:= t.tpe => unapply(prefix)
+      // Peel a method call that arrives via a single-argument adapter (typically an implicit conversion
+      // used to add extension methods, e.g. `componentExt(c).chain(...)` desugared from `c.chain(...)`).
+      // We require the adapter's converted argument to share its type with the whole expression -
+      // this preserves the rule's "shape-preserving chain" invariant: whatever the adapter does, the
+      // chain is still rooted in the same `Component[T]`, so the underlying `cached(...)` call is
+      // structurally reachable.
+      case UnwrapApply(Select(Apply(_, List(orig)), _)) if orig.tpe =:= t.tpe => unapply(orig)
       case _ => Some(t)
     }
   }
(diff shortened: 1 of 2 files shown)


halotukozak (approved):
LGTM, but remove the ai comments

mikegpl:
> LGTM, but remove the ai comments

Done

PR #865, opened 2026-06-01
[Scala 3] restore positioned.here + SourceInfo.here

halotukozak:
## Summary

Restores `annotation.positioned.here: Int` and `misc.SourceInfo.here: SourceInfo` from `???` stubs to working Scala 3 `inline def` + `scala.quoted` implementations.

- `positioned.here` returns `Position.ofMacroExpansion.start: Int` of the call site.
- `SourceInfo.here` returns a populated `SourceInfo` (filePath, fileName, offset, line, column, lineContent, enclosingSymbols) via `Position.ofMacroExpansion` + `Symbol.spliceOwner` owner-chain walk.
- Macro impls live inline next to their public `inline def` — no separate `commons.macros.*` package.
- `SourceInfo.apply()(implicit si: SourceInfo)` preserved (minimum-diff for downstream `summon[SourceInfo]` and `implicitly[SourceInfo]`).

## Tests

- New `PositionedTest` — asserts two adjacent `positioned.here` calls yield distinct positive `Int` offsets.
- Existing `SourceInfoTest` reactivated — pattern-matches expected `SourceInfo(...)` shape. Offset/column constants updated from 216/28 (Scala 2.13 macro semantics) to 205/17 (Scala 3 `Position.ofMacroExpansion` points at receiver start, not method name).

## MIGRATION.md

- Removes 2 backlog rows: `positioned.scala:12`, `SourceInfo.scala:28`.
- `Total tags: 155 -> 154`.
- Documents `sbt-ci-release` plugin disabled in § 5 (transitively pulls `sbt-git` whose JGit chokes on linked git worktrees with `NoWorkTreeException`). Restore effort: S — re-enable once releasing.

## Test plan

- [ ] `sbt commons-core/compile` exit 0
- [ ] `sbt 'commons-core/testOnly com.avsystem.commons.annotation.PositionedTest com.avsystem.commons.misc.SourceInfoTest'` exit 0
- [ ] `sbt scalafmtCheckAll scalafmtSbtCheck` exit 0


core/src/test/scala/com/avsystem/commons/annotation/PositionedTest.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class PositionedTest extends AnyFunSuite {
+  test("positioned.here yields distinct positive offsets at distinct call sites") {
+    val a = positioned.here
+    val b = positioned.here
+    assert(a > 0)
+    assert(b > 0)
+    assert(a != b)
```
MajronMan, line 15: I'd add a more specific test case, to see if all data is correct

halotukozak: sure, done

PR #881, opened 2026-06-02
[Scala 3] port ApplierUnapplier (Mirror-based)

halotukozak:
Restores `misc.ApplierUnapplier` family from Phase-1 `???` stubs to real Scala 3 typeclass implementations backed by `scala.deriving.Mirror.ProductOf`. Cribbed verbatim from fork `origin/master:core/src/main/scala-3/com/avsystem/commons/misc/ApplierUnapplier.scala`.

### What's restored

- `Applier[T <: Product]` — `Seq[Any] => T` via `Mirror.ProductOf[T] + Tuple.fromArray`.
- `Unapplier[T <: Product]` — `T => Seq[Any]` via `productIterator`.
- `ApplierUnapplier[T]` — combined.
- `ProductUnapplier[T]` / `ProductApplierUnapplier[T]` — concrete/abstract base classes for downstream extension.

All four use Scala 3 `given derived` with `@implicitNotFound` messages preserved.

### Tests

Restored `ApplierUnapplierTest` (was wrapped in Phase 1). Expanded coverage beyond the fork:

- Existing: `Empty`, `Single`, `Multiple`, `Varargs`, `Generic`, `Over22` (>22 params), `tuple`.
- New: nested case class, `Option` field, `assertDoesNotCompile` for sealed traits, `ProductApplierUnapplier` subclass usage, standalone `Applier`/`Unapplier` summoning.
- Removed: `ignore("custom")` stub — Mirror only derives for actual case classes, hand-written `apply`/`unapply` is out of scope.
- Bug fix: `Over22` default values `p14/p19/p20` were copy-paste typos (`"13"`/`"18"`/`"18"`).

12/12 tests green, 0 ignored.

### MIGRATION.md

`§3 core — misc ApplierUnapplier` entry; backlog rows for the three `materialize` Scala 2 macro stubs removed.

Slice 5.3 / Phase 5 leaf-feature-restoration.

### DIFF
--- a/core/src/main/scala/com/avsystem/commons/misc/ApplierUnapplier.scala
+++ b/core/src/main/scala/com/avsystem/commons/misc/ApplierUnapplier.scala
@@ -2,6 +2,7 @@ package com.avsystem.commons
 package misc
 
 import scala.annotation.implicitNotFound
+import scala.deriving.Mirror
 
 /** Typeclass which captures case class `apply` method in a raw form that takes untyped sequence of arguments.
   */
@@ -10,8 +11,8 @@ trait Applier[T] {
   def apply(rawValues: Seq[Any]): T
 }
 object Applier {
-  // TODO[scala3-port]: Applier.materialize (Scala 2 macro def) (L)
-  implicit def materialize[T]: Applier[T] = ???
+  given derived[T <: Product: Mirror.ProductOf as m]: Applier[T] = rawValues =>
+    m.fromTuple(Tuple.fromArray(rawValues.toArray).asInstanceOf[m.MirroredElemTypes])
 }
 
 /** Typeclass which captures case class `unapply`/`unapplySeq` method in a raw form that returns untyped sequence of
@@ -22,8 +23,7 @@ trait Unapplier[T] {
   def unapply(value: T): Seq[Any]
 }
 object Unapplier {
-  // TODO[scala3-port]: Unapplier.materialize (Scala 2 macro def) (L)
-  implicit def materialize[T]: Unapplier[T] = ???
+  given derived[T <: Product]: Unapplier[T] = value => IArraySeq.unsafeWrapArray(value.productIterator.toArray)
 }
 
 class ProductUnapplier[T <: Product] extends Unapplier[T] {
@@ -34,6 +34,8 @@ abstract class ProductApplierUnapplier[T <: Product] extends ProductUnapplier[T]
 @implicitNotFound("cannot materialize ApplierUnapplier: ${T} is not a case class or case class like type")
 trait ApplierUnapplier[T] extends Applier[T] with Unapplier[T]
 object ApplierUnapplier {
-  // TODO[scala3-port]: ApplierUnapplier.materialize (Scala 2 macro def) (L)
-  implicit def materialize[T]: ApplierUnapplier[T] = ???
+  given derived[T: {Applier as applier, Unapplier as unapplier}]: ApplierUnapplier[T] = new ApplierUnapplier[T] {
+    override def apply(rawValues: Seq[Any]): T = applier.apply(rawValues)
+    override def unapply(value: T): Seq[Any] = unapplier.unapply(value)
+  }
 }
--- a/core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
+++ b/core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
@@ -3,6 +3,15 @@ package com.avsystem.commons.misc
 import com.avsystem.commons
 import org.scalactic.source.Position
 import org.scalatest.funsuite.AnyFunSuite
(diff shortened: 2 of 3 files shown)


core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
```diff
@@ -3,6 +3,15 @@ package com.avsystem.commons.misc
 import com.avsystem.commons
 import org.scalactic.source.Position
 import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+case class Inner(s: String)
+case class Outer(inner: Inner, n: Int)
+case class WithOpt(o: Option[String], i: Int)
```
Andrzej-Swietek: WithOption - Opt is different thingy from Option

core/src/main/scala/com/avsystem/commons/misc/ApplierUnapplier.scala
```diff
@@ -34,6 +34,8 @@ abstract class ProductApplierUnapplier[T <: Product] extends ProductUnapplier[T]
 @implicitNotFound("cannot materialize ApplierUnapplier: ${T} is not a case class or case class like type")
 trait ApplierUnapplier[T] extends Applier[T] with Unapplier[T]
 object ApplierUnapplier {
-  // TODO[scala3-port]: ApplierUnapplier.materialize (Scala 2 macro def) (L)
-  implicit def materialize[T]: ApplierUnapplier[T] = ???
+  given derived[T: {Applier as applier, Unapplier as unapplier}]: ApplierUnapplier[T] = new ApplierUnapplier[T] {
```
Andrzej-Swietek, line 37: Named context bounds in complex form like `(T: {A as a, B as b})` is Scala 3.6+ feature - dont we want to  support LTS 3.3 ? FYL LTS 3.9 is said to loose compatibility with scala 2

core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
```diff
@@ -3,6 +3,15 @@ package com.avsystem.commons.misc
 import com.avsystem.commons
 import org.scalactic.source.Position
 import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+case class Inner(s: String)
```
Andrzej-Swietek: final ?

core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
```diff
@@ -3,6 +3,15 @@ package com.avsystem.commons.misc
...
+case class Outer(inner: Inner, n: Int)
+case class WithOpt(o: Option[String], i: Int)
+
+sealed trait Animal
```
Andrzej-Swietek: do we want those in packge misc scope or rather this test specific ?

core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
```diff
@@ -24,40 +33,29 @@ case class Over22(
...
-  def apply(str: String, i: Int): Custom = new Custom(str, i)
-  def unapply(custom: Custom): commons.Opt[(String, Int)] = Opt((custom.str, custom.i))
+given CustomProductApplierUnapplier: ProductApplierUnapplier[Multiple] = new {
+  override def apply(rawValues: Seq[Any]): Multiple =
+    Multiple(rawValues.head.asInstanceOf[String], rawValues(1).asInstanceOf[Int])
 }
 
-class ApplierUnapplierTest extends AnyFunSuite {
-  def roundtrip[T](
+class ApplierUnapplierTest extends AnyFunSuite with Matchers {
```
Andrzej-Swietek: final ? or specifically open otherwise

core/src/main/scala/com/avsystem/commons/misc/ApplierUnapplier.scala
halotukozak: we won't use it compatibility either way (because this library I macro-heavy). It would be better to have almost identical sources (in scala-2 and scala-3) with the copied content rather than one cross compiled, because we can use scala 3 features more freely, we can get used to scala 3 syntax, and our code will benefit and don't require another breaking change in the future (for example, we can replace implicit classes with extensions already, which have much better perf)

core/src/test/scala/com/avsystem/commons/misc/ApplierUnapplierTest.scala
halotukozak: if it comes to the inheritance, scala 3 classes are closed by default* (the * is really big). If it comes to the perf, it's a test file

Andrzej-Swietek (approved):

Andrzej-Swietek:
LGTM

PR #893, opened 2026-07-03
NativeJsonInput: treat object fields with undefined value as missing

### LINKED ISSUE #848: NativeJsonInput: treat object fields with `undefined` value as missing
### Problem

`NativeJsonInput` (scala-commons, com.avsystem.commons.serialization.nativejs) only treats a missing key in a `js.Dictionary` / `js.Object` as "no value". A key that is present but set to `undefined` is reported as a real
field, which then fails to decode — typically with a `ReadFailure` — instead of being treated as absent.

This is surprising in a JS context: in JavaScript, `{ foo: undefined }` and `{}` are effectively equivalent for property reads (`obj.foo === undefined` in both cases), and many JS-side producers emit `undefined` for optional
fields rather than omitting the key. As a result, payloads that round-trip through JS code (browser state, `js.Dictionary` built incrementally, libraries that set `obj.foo = undefined` for "no value") break deserialization
for fields that should be optional / have defaults.

### Reproduction

```scala
val dict = js.Dictionary[js.Any]("foo" -> js.undefined)
val input = new NativeJsonInput(dict.asInstanceOf[js.Any], NativeFormatOptions())
GenCodec.read[CaseClassWithOptionalFoo](input) // fails on "foo"
```

vs. the working case:

```scala
val dict = js.Dictionary[js.Any]() // "foo" not set at all
// reads fine — `foo` falls back to default / Opt.Empty
```

### Expected

`NativeJsonInput` should treat field: undefined the same as a missing field:

1. `ObjectInput.hasNext` / `nextField` should skip entries whose value is `js.isUndefined`.
2. `ObjectInput.peekField(name)` should return `Opt.Empty` when the key is present but its value is `undefined`, in addition to the key-not-present case.

sebaciv:
Resolves https://github.com/AVSystem/scala-commons/issues/848

### DIFF
--- a/core/js/src/main/scala/com/avsystem/commons/serialization/nativejs/NativeJsonInput.scala
+++ b/core/js/src/main/scala/com/avsystem/commons/serialization/nativejs/NativeJsonInput.scala
@@ -131,13 +131,16 @@ final class NativeJsonListInput(array: js.Array[js.Any], options: NativeFormatOp
 }
 
 final class NativeJsonObjectInput(dict: js.Dictionary[js.Any], options: NativeFormatOptions) extends ObjectInput {
-  private val it = dict.iterator
+  private val it = dict.iterator.filterNot({ case (_, value) => js.isUndefined(value) })
 
   override def hasNext: Boolean =
     it.hasNext
 
   override def peekField(name: String): Opt[FieldInput] =
-    if (dict.contains(name)) Opt(new NativeJsonFieldInput(name, dict(name), options)) else Opt.Empty
+    if (dict.contains(name) && !js.isUndefined(dict(name)))
+      Opt(new NativeJsonFieldInput(name, dict(name), options))
+    else
+      Opt.Empty
 
   override def nextField(): FieldInput = {
     val (key, value) = it.next()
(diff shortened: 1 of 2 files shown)


core/js/src/main/scala/com/avsystem/commons/serialization/nativejs/NativeJsonInput.scala
```diff
@@ -131,13 +131,16 @@ final class NativeJsonListInput(array: js.Array[js.Any], options: NativeFormatOp
...
 
   override def hasNext: Boolean =
     it.hasNext
 
   override def peekField(name: String): Opt[FieldInput] =
-    if (dict.contains(name)) Opt(new NativeJsonFieldInput(name, dict(name), options)) else Opt.Empty
+    if (dict.contains(name) && !js.isUndefined(dict(name)))
+      Opt(new NativeJsonFieldInput(name, dict(name), options))
+    else
+      Opt.Empty
```
Andrzej-Swietek: peekField  now does redundant lookups in the hit path ? `dict.contains(name)` then `dict(name)` in the `isUndefined` check
```scala
override def peekField(name: String): Opt[FieldInput] =
  dict.get(name).filterNot(js.isUndefined).map(new NativeJsonFieldInput(name, _, options)).toOpt
```

Andrzej-Swietek (approved):

core/js/src/main/scala/com/avsystem/commons/serialization/nativejs/NativeJsonInput.scala
sebaciv: Changed to WrappedDictionary.get + more tests

Andrzej-Swietek (approved):

